PR #18, opened 2017-10-11
first notebook and cheyenne launch scripts

jhamman:
This PR has the following things in it:

- Example notebook that @mrocklin and I put together analyzing some ensemble met data
- Some nice wrapper scripts for launching a dask-distributed cluster and jupyter hub server on cheyenne, Usage:
    ```Bash
    cd utilities/cheyenne/
    ./launch-dask.sh $n_worker_nodes
    ```
- A conda environment file with a more complete pangeo spec
- Updates to gitignore/readme files

*Note: I'm not sure we need/want to keep the wrapper scripts in this repo but I noticed copies were beginning to propagate around Cheyenne so I think its good to put them in the repo as a means of sharing them for now.*

cc @rabernat for the notebook
cc @mrocklin for the wrapper scripts


rabernat:
Nitpicking the text is definitely not a high priority right now...but the first word of the notebook just can't be "Analsys". 😬  

A few other minor suggestions, considering that this will probably become a de-facto template
- It would be nice to also see the xarray repr of the dataset (in addition to `ds.info()`)
- Explain the line isel statement in `t_mean = ds['t_mean'].isel(ensemble=slice(0, 25))`. (I assume it's because you don't have enough memory for the whole ensemble?)

Other than that, it looks great. It's a simple example, hopefully understandable by everyone involved..

jhamman:
> Nitpicking the text is definitely not a high priority right now...but the first word of the notebook just can't be "Analsys". 😬

I think I'll fix that. Yikes!

Happy to spend some more time cleaning up the notebook.

utilities/cheyenne/setup-jlab.py
```diff
@@ -0,0 +1,75 @@
...
+@click.option('--hostname', default='cheyenne.ucar.edu', show_default=True,
+              help='public facing hostname')
+@click.option('--log_level', default='INFO', show_default=True,
+              help='logging level, useful for debugging')
+def cli(scheduler_file, jlab_port, dash_port, notebook_dir,
+        hostname, log_level):
+    logger = get_logger(log_level)
+
+    logger.info('getting client with scheduler file: %s' % scheduler_file)
+    client = Client(scheduler_file=scheduler_file)
```
mrocklin: We could also put a large `timeout=20` here rather than use the `sleep 10` in the launching script

utilities/cheyenne/setup-jlab.py
```diff
@@ -0,0 +1,75 @@
...
+    logger.info('host is %s' % host)
+
+    logger.info('Starting jupyter lab on host')
+    client.run_on_scheduler(start_jlab, host=host, port=jlab_port,
+                            notebook_dir=notebook_dir)
+    logger.debug('Done.')
+
+    print('Running the following command from your local hostname:')
+    print(f'ssh -N -L {jlab_port}:{host}:{jlab_port} '
+          f'-L {dash_port}:{host}:8787 {hostname}')
```
mrocklin: One of the comments we got was that this didn't work well for people when they had a different username locally than on the cluster.  We might want to add `os.environ['USER']` somewhere in here as well.

utilities/cheyenne/launch-dask.sh
```diff
@@ -0,0 +1,35 @@
...
+    sleep 1
+done
+
+sleep 10 # give the scheduler a bit of time to get started
+
+source activate pangeo
+
+echo "Setting up Jupyter Lab"
+./setup-jlab.py --log_level=DEBUG --jlab_port=8877 --dash_port=8878 \
+    --notebook_dir /glade/p/work/jhamman/pangeo/src/pangeo/notebooks/ \
```
mrocklin: You have your username in here

jhamman:
@mrocklin and @rabernat - thanks for the comments. All have been addressed.

PR #27, opened 2017-11-07
changes to default cheyenne cluster specs

jhamman:
Working with @mrocklin, we discovered a few problems with the default cluster specs on Cheyenne. This, combined with https://github.com/dask/distributed/pull/1520 should fix those problems. 

xref: https://github.com/dask/distributed/issues/1469

### DIFF
--- a/utilities/cheyenne/launch-dask-scheduler.sh
+++ b/utilities/cheyenne/launch-dask-scheduler.sh
@@ -2,7 +2,7 @@
 #PBS -N dask-scheduler
 #PBS -q economy
 #PBS -A UCLB0022
-#PBS -l select=1:ncpus=36:mpiprocs=6:ompthreads=6
+#PBS -l select=1:ncpus=36:mpiprocs=9:ompthreads=4:mem=109GB
 #PBS -l walltime=11:59:00
 #PBS -j oe
 #PBS -m abe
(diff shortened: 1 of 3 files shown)


utilities/cheyenne/launch-dask-scheduler.sh
```diff
@@ -18,5 +18,8 @@ source activate pangeo
 
 SCHEDULER=/glade/scratch/$USER/scheduler.json
 rm -f $SCHEDULER
-mpirun --np 6 dask-mpi --nthreads 6 --memory-limit 22e9 --interface ib0 \
-    --local-directory /glade/scratch/$USER --scheduler-file=$SCHEDULER
+mpirun --np 9 dask-mpi --nthreads 4 \
+    --memory-limit 12e9 \
```
mrocklin: The 109GB limit is genuine?  Should we be more conservative here?  (I'm hoping the answer is "no")

jhamman: These nodes have 128GB total RAM and 109GB available. 9x12=108 so I think we'll be okay, provided that the nannies never let a task run over 12GB.

davidedelvento: See https://github.com/matthewsucar/userspace_oom for details on the nannies.
My understanding (not guaranteed to be correct) is that it cares only about total utilization.

PR #50, opened 2017-12-12
launch-dask.py script

jedwards4b:
launch-dask.py combines the functions of launch-dask.sh and setup_jlab.py into a single
script and provides command line options for the walltime and project number.    More could be done to make this a flexible and reliable tool, but this is a start.


davidedelvento:
@kmpaul @mrocklin Folks, could you please take a look at this?

mrocklin:
I don't have any strong thoughts on this.  The person with the most recent experience here is probably @jhamman .

I'm curious what activities led to these changes. Were you running into problems with prior approaches @jedwards4b ?

jedwards4b:
Yes, there are lots of hard coded parameters in those scripts - wallclock time, project id, etc. 
I've also reduced the number of steps required.

mrocklin:
Was this done in support of a particular group or just because it seemed like the right thing to do?  I'm trying to get a sense for who is using this work.

rabernat:
Note that I also have custom launch scripts for our habanero cluster
https://github.com/pangeo-data/pangeo/tree/master/utilities/habanero
and TACC wrangler:
https://github.com/pangeo-data/pangeo/tree/master/utilities/wrangler

I agree that standardizing the launch script is a good goal. Are we prepared to support this on multiple different platforms? Some of these platforms use PBS, other SLURM.

mrocklin:
My general approach in situations like this is to use special cased
solutions until they become burdensome and generalize only when
demonstrably valuable.  It's entirely possible that we'll find out new and
better solutions to these sorts of issues in the meantime.

mrocklin:
That's just a general note though, not a statement about what is
appropriate in this situation (you all seem closer to the ground at this
point than I am)

utilities/cheyenne/launch-dask.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python
+from __future__ import absolute_import
+from __future__ import print_function
+
+import os, sys, argparse, subprocess, re, time, logging, socket
```
jhamman: A few general comments about python style (PEP8 nits): 

- one import per line
- two lines between functions
- spaces after commas, etc.

utilities/cheyenne/launch-dask.py
```diff
@@ -0,0 +1,131 @@
...
+from dask.distributed import Client
+logger = logging.getLogger(__name__)
+
+def parse_command_line(args, description):
```
jhamman: The `setup_jlab` script uses `click` for the parsing the command line arguments. I think you'll find that is an easier/less verbose way to build command line tools.

utilities/cheyenne/launch-dask.py
```diff
@@ -0,0 +1,131 @@
...
+                        help="Set the wallclock limit for all nodes. ")
+
+    parser.add_argument("--workdir",
+                        default=os.path.join("/glade","scratch",os.environ.get("USER")),
+                        help="Set the working diirectory")
+
+    parser.add_argument("--notebookdir",
+                        default=os.path.join("/glade","p","work",os.environ.get("USER")),
+                        help="Set the working diirectory")
+
```
jhamman: It would also be good to configure which queue these jobs land on, as well as the breakdown of processes/threads. Also the notebook and dask-dashboard ports should be configurable.

utilities/cheyenne/launch-dask.py
```diff
@@ -0,0 +1,131 @@
...
+    search_jobid = re.search("^(\S+)$", output.decode('utf-8'))
+    jobid = search_jobid.group(1)
+    return jobid
+
+def start_jlab(dask_scheduler, host=None, port='8888', notebook_dir=''):
+    cmd = ['jupyter', 'lab', '--ip', host,
+           '--no-browser', '--port', port,
+           '--notebook-dir', notebook_dir]
+#    logger.debug("cmd={}".format(cmd))
```
jhamman: remove this commented out line.

utilities/cheyenne/launch-dask.py
```diff
@@ -0,0 +1,131 @@
...
+    print('Run the following command from your local machine:')
+    print(f'ssh -N -L {jlab_port}:{host}:{jlab_port} '
+          f'-L {dash_port}:{host}:8787 {user}@{hostname}')
+    print('Then open the following URLs:')
+    print(f'\tJupyter lab: http://localhost:{jlab_port}')
+    print(f'\tDask dashboard: http://localhost:{dash_port}', flush=True)
+
+def _main_func(args, description):
+    project, workers, walltime, notebookdir, workdir = parse_command_line(args, description)
+    global logger
```
jhamman: I don't think you need this `global` statement.

utilities/cheyenne/launch-dask.py
```diff
@@ -0,0 +1,131 @@
...
+    logger = get_logger("DEBUG")
+
+    jobids = []
+    jobids.append( launch_job("launch-dask-scheduler.sh", project, walltime))
```
jhamman: It may be worth doing away with these shell scripts all together. It should be possible to represent the full workflow from with a single command line execution. It actually may be a nice way to make this much more portable.

utilities/cheyenne/launch-dask.py
```diff
@@ -0,0 +1,131 @@
...
+            output, errput = proc.communicate()
+            output = output.decode('utf-8')
+            stat = proc.wait()
+            if ' R ' in output:
+                logger.info(" jobid {} started".format(jobid))
+                time.sleep(5) # make sure scheduler file is written
+                wait = False
+            if ' Q ' in output:
+                logger.info(" jobid {} in queue".format(jobid))
+                time.sleep(5)
```
jhamman: We only need to block until the scheduler job is running. The rest of the worker jobs can come later.

utilities/cheyenne/launch-dask.py
```diff
@@ -0,0 +1,131 @@
...
+
+    setup_jlab(jlab_port='8877', dash_port='8878', notebook_dir=notebookdir,
+               hostname="cheyenne.ucar.edu",
+               scheduler_file=os.path.join(workdir,"scheduler.json"))
```
jhamman: I think I said this above, we should be able to modify the jlab/dash ports via the command line tool.

utilities/cheyenne/launch-dask.py
jedwards4b: I prefer argparse as it is standard python and I am familiar with it.

jhamman: okay, that is fine. Just so you know, the dask command line tools (`dask-mpi`) use click so that is a dependency this tool already has.

mrocklin: In general I think that conventions like this should generally be decided by whoever plans to maintain the code long term.  It's not clear who that is in this situation.  Are you planning to manage bug reports and feature requests on this code long term @jedwards4b ?  Or is future development likely to be left to the Dask/XArray developers?  If the latter then I think that we should stick with `click`.

mrocklin: (I'm not saying -1 here though.  I'm not heavily involved in this at the moment.  Happy to go with @jhamman 's decision)

jedwards4b:
@mrocklin - if you are reinventing this wheel in pr #56 why not mention it in your review so that we don't duplicate effort?

mrocklin:
> @mrocklin - if you are reinventing this wheel in pr #56 why not mention it in your review so that we don't duplicate effort?

I don't believe that #56 significantly overlaps with this functionality.  I apologize if the lack of communication has troubled you.  I'd be happy to sync up any time if you're interested.  I'm curious what you all are working on.

jedwards4b:
I see a lot of overlap but okay, no worries.

mrocklin:
Do you have any interest in a video chat at some point?  If you're going to be working on this effort it would be good to ensure that we're all aware of each others' actions.

jedwards4b:
Eventually yes, but it may be a bit premature - I have a lot to learn.

mrocklin:
> Eventually yes, but it may be a bit premature - I have a lot to learn.

Well, if you're planning to do work in this area then it would be good to ensure that you're talking to someone and that that work is in line with other efforts.  This could be me or someone like @jhamman if you want to keep things in-house.  If as you say you have a lot to learn then this is especially true.

jedwards4b:
What's the difference between dask-worker and dask-mpi?

mrocklin:
dask-worker is the standard way to launch a single dask worker.  It is typically used with many different cluster management solutions.  `dask-mpi` is a convenience program specifically for launching many dask workers using MPI.  The relevant documentation page is http://dask.pydata.org/en/latest/setup.html

mrocklin:
If you're going to work in this space you should probably read through the following pages in particular:

-  http://dask.pydata.org/en/latest/setup/cli.html
- http://dask.pydata.org/en/latest/setup/hpc.html

jedwards4b:
So the only difference is in how the workers are launched?   I'm asking because it seems iike the way to move forward with this PR is to merge in #56 and then import PBSCluster in launch-dask.py - would you agree?

mrocklin:
I think it depends on what our objectives are.  For example if people want to manage PBS scripts on their own then #56 is not a good solution for them and what you have going here is probably nicer.  I'm not sure what motivated the work in this PR, so I don't have strong thoughts on the right way forward.

As an example #56 was specifically designed to enable adaptive deployments (where the number of active workers changes based on load) launched from a notebook as was requested by @jhamman and motivated from his interactions trying to get others in his group to use this work who were a bit daunted by our current setup process.  It was not intended to replace the current process.

jedwards4b:
Who really wants to manage PBS scripts on their own?  What are the drawbacks to replacing the current (very awkward IMHO) process?

mrocklin:
> Who really wants to manage PBS scripts on their own? 

For the people I've worked with recently they seem to be familiar with PBS scripts and appreciate that they know how to extend things without having to interact with us.  They prefer not to be boxed in by whatever I thought was best at the time.

> What are the drawbacks to replacing the current (very awkward IMHO) process?

I have no objection to replacing the current process.  You asked about the best way to move forward and I was responding saying that I don't have a good objective function here.  Typically I motivate development by user interactions.  You may have a better understanding of what typical users want than I do when setting things up on PBS systems.

jhamman:
@jedwards4b - I pushed some minor changes here (to your branch). The main change is moving the pbs templates into the python script so they don't have to travel together. There is plenty more generalization that could occur here but this seems to be hitting the leading offenders. I'd be interested to hear how this works for you.

jedwards4b:
I much prefer the method using the pangeo class and think we should focus there

jhamman:
@jedwards4b - I agree. I had these changes sitting in my workspace so I decided it was better to push them than to let them get lost. If there isn't any interest in finishing this, I'm happy to see it closed.

PR #54, opened 2017-12-18
setup launch scripts for geyser/caldera

jhamman:
Added scripts for launching jupyer/dask on slurm for the geyser/caldera cluster. This doesn't work quite yet but I'm putting it up anyways.

### DIFF
--- a/utilities/geyser_caldera/launch-dask-scheduler.sh
+++ b/utilities/geyser_caldera/launch-dask-scheduler.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+#SBATCH -C geyser
+#SBATCH -J dask-scheduler
+#SBATCH -n 10
+#SBATCH --mem-per-cpu=50G
+#SBATCH -t 11:59:00
+#SBATCH -A UCLB0022
+#SBATCH -p dav
+#SBATCH -e dask-scheduler.err.%J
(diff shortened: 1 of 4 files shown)


jhamman:
@davidedelvento - care to take a peak at the slurm components here?

jedwards4b:
I think that all of these individual shell scripts with multiple hard coded values are a bad idea.  Would you consider PR #50 and the possibility of merging the cheyenne/geyser/caldera scripts into a smaller set of more flexible tools?

jhamman:
@jedwards4b - certainly. I'll review your PR now. I think this gets to a broader question of if and to what extent these tools will be used and supported on Cheyenne. Is CISL considering making this sorts of tools available to a broader user group?

davidedelvento:
@jhamman sure, when they are production ready. That is the reason why I got involved in the first place!

jhamman:
@davidedelvento - great! How do should we define production ready?

davidedelvento:
How about when @kmpaul deems it so? Maybe under the threaten than he'll get the user request for assistance? :-)

jhamman:
I'm going to close this for now. CISL is still working on enabling port forwarding on Geyser/Caldera and until that happens, this is not worth pursuing.

PR #56, opened 2017-12-22
Add draft of PBSCluster

mrocklin:
This provides a Pythonic way to launch Dask clusters on a PBS cluster
from Python.

This also works with adaptive clusters

Examples
--------

```python
>>> from pangeo import PBSCluster
>>> cluster = PBSCluster(project='...')
>>> cluster.start_workers(10)  # this may take a few seconds to launch

>>> from dask.distributed import Client
>>> client = Client(cluster)

This also works with adaptive clusters.  This automatically launches and
kill workers based on load.

>>> cluster.adapt()
```

### DIFF
--- a/pangeo/pbs.py
+++ b/pangeo/pbs.py
@@ -1,0 +1,163 @@
+from contextlib import contextmanager
+import logging
+import os
+import socket
+import subprocess
+import sys
+
+from distributed import LocalCluster
+from distributed.utils import tmpfile, get_ip_interface
+
+template = """
+#!/bin/bash
+#PBS -N %(name)s
+#PBS -q %(queue)s
+#PBS -A %(project)s
+#PBS -l select=1:ncpus=%(threads_per_worker)d:mpiprocs=1:ompthreads=1
+#PBS -l walltime=%(walltime)s
+#PBS -j oe
+#PBS -m abe
+
+%(base_path)s/dask-worker %(scheduler)s --nthreads %(threads_per_worker)d --memory-limit %(memory)s --name $PBS_JOBNAME-$PBS_JOBID-$PBS_TASKNUM %(extra)s
+"""
+
+
+logger = logging.getLogger(__name__)
+
+
+dirname = os.path.dirname(sys.executable)
+
+
+class PBSCluster(object):
+    """ Launch Dask on a PBS cluster
+
+    Examples
+    --------
+    >>> from pangeo import PBSCluster
+    >>> cluster = PBSCluster(project='...')
+    >>> cluster.start_workers(10)  # this may take a few seconds to launch
+
+    >>> from dask.distributed import Client
+    >>> client = Client(cluster)
+
+    This also works with adaptive clusters.  This automatically launches and
+    kill workers based on load.
+
+    >>> from distributed.deploy import Adaptive
+    >>> adapt = Adaptive(cluster.cluster.scheduler, cluster)
+    """
+    def __init__(self,
+                 name='dask',
+                 q='regular',
+                 project='UCLB0022',  # is there an environment variable I can use?
+                 threads_per_worker=6,
+                 memory='7GB',
+                 walltime='00:30:00',
+                 interface=None,
+                 extra='',
+                 scheduler_file=os.path.expanduser('~/scheduler.json'),
+                 **kwargs):
+        if interface:
+            host = get_ip_interface(interface)
+            extra += '--interface ' + interface
+        else:
+            host = socket.gethostname()
+        self.cluster = LocalCluster(n_workers=0, ip=host, **kwargs)
+        self.config = {'name': name,
+                       'queue': q,
+                       'project': project,
+                       'threads_per_worker': threads_per_worker,
+                       'walltime': walltime,
+                       'scheduler': self.cluster.scheduler.address,
+                       'base_path': dirname,
+                       'memory': memory.replace(' ', ''),
+                       'extra': extra}
+        self.jobs = set()
+
+    def job_script(self):
+        return template % self.config
+
+    @contextmanager
+    def job_file(self):
+        """ Write job submission script to temporary file """
+        with tmpfile(extension='sh') as fn:
+            with open(fn, 'w') as f:
+                f.write(self.job_script())
+
+            yield fn
+
+    def start_workers(self, n=1):
+        """ Start workers and point them to our local scheduler """
+        with self.job_script() as fn:
+            outs = self._calls([['qsub', fn]] * n)
+            jobs = [out.decode().split('.')[0] for out in outs]
+        self.jobs.update(jobs)
+        return jobs
+
+    @property
+    def scheduler_address(self):
+        return self.cluster.scheduler_address
+
+    def _calls(self, cmds):
+        """ Call a command using subprocess.communicate
+
+        This centralzies calls out to the command line, providing consistent
+        outputs, logging, and an opportunity to go asynchronous in the future
+
+        Parameters
+        ----------
+        cmd: List(List(str))
+            A list of commands, each of which is a list of strings to hand to
+            subprocess.communicate
+
+        Examples
+        --------
+        >>> self._calls([['ls'], ['ls', '/foo']])
+
+        Returns
+        -------
+        The stdout result as a string
+        Also logs any stderr information
+        """
+        logger.debug("Submitting the following calls to command line")
+        for cmd in cmds:
+            logger.debug(' '.join(cmd))
+        procs = [subprocess.Popen(cmd,
+                                  stdout=subprocess.PIPE,
+                                  stderr=subprocess.PIPE)
+                 for cmd in cmds]
+
+        result = []
+        for proc in procs:
+            out, err = proc.communicate()
+            if err:
+                logger.error(err.decode())
+            result.append(out)
+        return result
+
+    def _call(self, cmd):
+        """ Singular version of _calls """
+        return self._calls([cmd])[0]
+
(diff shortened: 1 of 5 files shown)


pangeo/pbs.py
```diff
@@ -0,0 +1,163 @@
...
+    This also works with adaptive clusters.  This automatically launches and
+    kill workers based on load.
+
+    >>> from distributed.deploy import Adaptive
+    >>> adapt = Adaptive(cluster.cluster.scheduler, cluster)
+    """
+    def __init__(self,
+                 name='dask',
+                 q='regular',
+                 project='UCLB0022',  # is there an environment variable I can use?
```
mrocklin: Is there a way to obtain a default project for a particular user?  @davidedelvento

mrocklin:
@jhamman this may interest you.  If you get a chance to test drive this I'd appreciate it.  Please feel free to push to this branch or make comments as you like.

jhamman:
Thanks @mrocklin - giving it a go now.

pangeo/pbs.py
davidedelvento: PBS does not have a concept of default project, however NCAR added it for users who wish to set it up for their own reasons, by using the `$PBS_ACCOUNT` environmental variable.

Some users set it, and it'd be great if you'd use it for them, but many users do not set it. So you need to cover for when it's missing. In this case, you can pick the first project they have or one at random. See how we (== the group I work in, aka CSG) implemented this logic in `/glade/u/apps/ch/opt/usr/bin/qcmd` and `/glade/u/apps/ch/opt/usr/bin/qinteractive` (they are bash scripts, IIRC). Both of them (IIRC) rely on `/ncar/opt/sam/sam_validate.py` which is out of the control of CSG (as you can see it's owned by root, which we don't have). Also, you can see this thing is not truly made for importing, but I could see if I find who maintains it and put you in touch with them if you prefer to do use a stable API vs parsing text as bash does.

jhamman:
@mrocklin - I made a few edits. I was able to get the PBSCluster to work but was not successful in getting the adaptive cluster to scale up/down. I'll give it another try next week.

mrocklin:
Yeah, I think that clean scale down depends on https://github.com/dask/distributed/pull/1659

pangeo/pbs.py
```diff
@@ -0,0 +1,208 @@
...
+        name : str
+            Name of worker jobs. Passed to `$PBS -N` option.
+        queue : str
+            Destination queue for each worker job. Passed to `#PBS -q` option.
+        project : str
+            Accounting string associated with each worker job. Passed to
+            `#PBS -A` option.
+        resource_spec : str
+            Request resources and specify job placement. Passed to `#PBS -l`
+            option.
```
mrocklin: How should we reconcile this argument and the `memory` and `threads_per_worker` argument?

Also, I'm curious, what was the reason for this change?  I would not be surprised to learn that the previous solution was restrictive in some way.

Also, why the multiple processes per job change?  Is this nicer than having many small jobs?  Would array jobs be more appropriate?  From a cleanup perspective it's nice for the `PBSCluster` to have a handle on each individual worker separately.

mrocklin: Short term I've reverted this and gone back to only supporting `--nthreads`.  It is difficult to use `--nprocs`, especially if we want to do adaptive deployments.

jhamman: Yes, motivation for this change was:

1) The resource specification should be flexible enough to move between different machines that use PBS. While `#PBS -l select=1:ncpus=$ncpus:mem=$mem` works on cheyenne, this is not going to work everywhere. I figured it is easy enough to allow users to specify this line.
2) From the [Cheyenne docs](https://www2.cisl.ucar.edu/resources/computational-systems/cheyenne/quick-start-cheyenne):
    > Most of the Cheyenne batch queues are for exclusive use, and jobs are charged for all 36 cores on each node that is used. Jobs in the shared use "share" queue are charged only for the cores used.

    This is why I had setup the PBSCluster to use multiple workers per job. Alternatively, we could a) try using the shared queue, b) accept paying for cores we're not using, or c) use 36 threads per worker (good for cpu bound computations but not IO bound tasks).

mrocklin:
Ok, I'll take another crack at this

pangeo/pbs.py
davidedelvento: I suggest avoid option a) at least by default, since sharing node with others can cause a lot of troubles (at least at the moment). Mostly, but not only, because users have workarounds that allow them to use more memory, CPU and network than their share, overloading the node (we plan to deploy a more aggressive watchdog, angrily biting whoever spills over their share, but it's not in place and won't be for a while). Option c) seems best. If you are IO bound, oh well. How many IO bound jobs are a user going to run? For how long total compared to their own total allocation? I suspect for most folks it'd be irrelevant cost. And for the few people it isn't, they could use the non-default a) option and deal with the issues.

mrocklin: I also like option (c) from a Dask perspective, however I'm also sensitive to IO bound concerns.  I suspect that our computations are more I/O bound than most that are run on Cheyenne.

@jhamman are you mainly talking about disk I/O or inter-worker network communication?  If disk I/O then do we know what the limiting factor is here?  Is it compression libraries that don't release the GIL?

jhamman: @mrocklin - yes, disk I/O. I suspect it will be difficult to clearly define the limiting factor on disk I/O. It will greatly depend on both how the data is stored and the chunk configuration. One thing we know for sure is that more processes certainly helps.

mrocklin:
One thing to do here eventually is to load some data with a many-threaded
worker and then check the distributed profiler.  That'll probably give us a
sense about where the hotspots are.

mrocklin:
@jhamman I believe that I've handled the multi-processing issues.  This depends on a development branch in https://github.com/dask/distributed/pull/1632

```python
from pangeo import PBSCluster
cluster = PBSCluster(project='...')
cluster.adapt()

from dask.distributed import Client
client = Client(cluster)
```

Then you'll get nine processes with four threads each by default.  The cluster will scale up and down by full nodes.  We group worker processes together by node before we determine which to clean up.

jhamman:
Thanks @mrocklin - I'll give it a try and report back.

pangeo/pbs.py
```diff
@@ -0,0 +1,236 @@
...
+
+#PBS -N %(name)s
+#PBS -q %(queue)s
+#PBS -A %(project)s
+#PBS -l %(resource_spec)s
+#PBS -l walltime=%(walltime)s
+#PBS -j oe
+#PBS -m abe
+
+%(base_path)s/dask-worker %(scheduler)s \
```
jedwards4b: Where is the associated dask-worker script?

mrocklin: This is installed when you install dask.distributed

jedwards4b:
```self = <pangeo.pbs.PBSCluster object at 0x7fffc6685828>

    def adapt(self):
        """ Start up an Adaptive deployment if not already started
    
            This makes the cluster request resources in accordance to current
            demand on the scheduler """
        from distributed.deploy import Adaptive
        if self._adaptive:
            return
        else:
            self._adaptive = Adaptive(self.scheduler, self, startup_cost=5,
>                                     key=lambda ws: ws.host)
E           TypeError: __init__() got an unexpected keyword argument 'key'

/glade/u/home/jedwards/sandboxes/pangeo/pangeo/pbs.py:236

**Comment by mrocklin** [comment]
> This depends on a development branch in dask/distributed#1632

**Comment by jedwards4b** [comment]
Ah - how do I correctly install that development branch?  I did this:
```

mrocklin:
Something like the following would probably work:

```
source activate pangeo  # use the pangeo environment
pip install git+https://github.com/asford/distributed.git@grouped_adaptive
```

But as always this depends on how you've set up your environment.

jedwards4b:
Thanks, I manually linked and that fixed the problem, tests are working now.

mrocklin:
Centralizing conversation from https://github.com/mrocklin/pangeo/pull/1 to here.

@jedwards4b asks 

> Are you saying that PBSCluster would allow you to start dask scheduler and workers from the notebook session directly? 

Yes, that is the intended user experience behind this PR.

> Maybe it would help if you provided an example notebook?

The top comment of this PR provides a minimal example.

jedwards4b:
Your introductory comment was ```This provides a Pythonic way to launch Dask clusters on a PBS cluster
from Python.``` now you are saying that this is a way to launch a dask cluster on a PBS system through a notebook interface and is not an appropriate method outside a notebook.    

I think I'm beginning to understand, thanks.

mrocklin:
I apologize if my communication has frustrated you.  As stated in https://github.com/pangeo-data/pangeo/pull/50#issuecomment-354322855 this might not be the right path for you.  I don't know though because I'm not sure what it is that you're trying to accomplish.

pangeo/pbs.py
```diff
@@ -0,0 +1,236 @@
...
+    --nthreads %(threads_per_worker)d \
+    --nprocs %(processes)s \
+    --memory-limit %(memory)s \
+    --name %(name)s-%(n)d \
+     %(extra)s
+""".lstrip()
```
jhamman: In the future, we may want to make this template a attribute on the PBSCluster object. For use on Cheyenne, this will certainly work so I'm not suggesting changing this until we want to be entirely machine agnostic.

mrocklin:
Sure.  Merging.  Was just waiting on feedback.

pangeo/pbs.py
mrocklin: Agreed on both counts.

PR #57, opened 2017-12-26
Add basic JupyterHub + Dask deployment

mrocklin:
This is a vanilla JupyterHub deployment alongside a vanilla Dask deployment .  It includes instructions and helm config files to set things up.


gce/jupyter-config.yaml
```diff
@@ -0,0 +1,24 @@
+proxy:
+  secretToken: SECRET
+
+singleuser:
+  image:
+    name: daskdev/dask-notebook
+    tag: latest
+  extraEnv:
+    EXTRA_PIP_PACKAGES: gcsfs xarray
```
jhamman: How hard would it be to put xarray-master in here instead? Also, if we can the optional xarray dependencies (scipy/netcdf4/zarr), that will allow us to start testing a few workflows.

mrocklin: We can anything that is pip or conda installable.  Are there particular branches that you'd like us to target?

mrocklin: I would also be quite happy to walk you through how to manipulate the cluster if you're interested.  Changing this stuff is surprisingly easy.

mrocklin: Conda installing scipy/netcdf4 each time is taking a while.  I'm building and pushing new docker images.

mrocklin: I've created new docker images and included the files in this PR.  We now have scipy and netCDF4 installed in the image and pip install gcsfs, xarray, and zarr from git master on each image load.

mrocklin:
> @mrocklin - What do you want to do here? Should we merge this and then incrementally add functionality to these configs? Autoscaling would be an obvious feature to work in but that doesn't have to happen immediately.

I'm happy to wait before merging.

mrocklin:
OK, for fuse it looks like the docker container needs to be run with elevated priveleges.  Locally I do the following:

    docker run -it --device /dev/fuse --cap-add SYS_ADMIN --privileged daskdev/pangeo-notebook

How do I specify flags like ` --device /dev/fuse --cap-add SYS_ADMIN --privileged` in the `config.yaml` file?  cc @yuvipanda

gce/setup.sh
```diff
@@ -0,0 +1,30 @@
+# Start cluster on Google cloud
+gcloud container clusters create pangeo --num-nodes=10 --machine-type=n1-standard-2 --zone=us-central1-b  --cluster-version=1.8.4-gke.1
+gcloud container clusters get-credentials pangeo --zone us-central1-b --project pangeo-181919
```
rabernat: At some point, using preemptible VMs might save us lots of money
https://cloud.google.com/kubernetes-engine/docs/concepts/preemptible-vm

gce/setup.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Set up Kubernetes
+kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=mrocklinWgmail.com
```
rabernat: typo in email address?

gce/notebook/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM daskdev/dask-notebook
+    
+RUN conda install --yes -c conda-forge \
+    xarray \
+    netcdf4 \
+    scipy \
+    && conda clean -tipsy
```
rabernat: Maybe we should be using a standard environment file here?

gce/setup.sh
mrocklin: I agree.  There is some delay in launching new nodes, but generally I agree that long term this is a good strategy.  Short term I'm just leaving a few nodes on for a while.  I'm judging that the cost in expense is worth it to reduce friction when getting started.  Please let me know if you feel differently and I can adjust.

gce/setup.sh
mrocklin: Thanks

gce/notebook/Dockerfile
mrocklin: Sure, I would be fine with that

gce/jupyter-config.yaml
```diff
@@ -0,0 +1,64 @@
...
+#       contactEmail: mrocklin@gmail.com
+
+auth:
+ type: github
+ github:
+   clientId: "2cb5e09d5733ff2e6ae3"
+   clientSecret: SECRET
+   callbackUrl: "http://pangeo.pydata.org/hub/oauth_callback"
+
+admin:
```
yuvipanda: This needs to be under 'auth' for it to take effect.

gce/setup.sh
rabernat: This is good for now.

gce/notebook/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+                git+https://github.com/jhamman/xarray@fix/zarr_set_attrs \
+                git+https://github.com/martindurant/gcsfs.git@fuse \
+                fusepy
+
+USER root
+RUN mkdir /gcs && chown -R $NB_USER /gcs
+COPY prepare.sh /usr/bin/prepare.sh
+RUN chmod +x /usr/bin/prepare.sh
+RUN mkdir /home/$NB_USER/examples && chown -R $NB_USER /home/$NB_USER/examples
+COPY examples/ /home/$NB_USER/examples
```
mrocklin: @yuvipanda will this work or will the user's home directory be overridden by jupyterhub?

yuvipanda:
It will be overwritten. You can put them in a non home directory and copy
them post start. Or you can turn off persistent storage completely and that
will make these show up.

mrocklin:
Persistent storage seems pretty valuable.  I'll copy them over in the
startup script.  Thanks!

rabernat:
I am trying out the GCP pangeo environment on a larger dataset. Here is my code

```python
import numpy as np
import xarray as xr
import matplotlib.pyplot as plt
from dask.distributed import Client, progress

from daskernetes import KubeCluster
cluster = KubeCluster(n_workers=10)
cluster
# then I scale to 100 workers

client = Client(cluster)
client

import gcsfs
gcsmap = gcsfs.mapping.GCSMap('pangeo-data-private/CM2.6/control/ocean_minibling_field/')
ds = xr.open_zarr(gcsmap, decode_cf=False)

ds.salt.mean().load()
```

Despite seeing the 100 workers in the dashboard, the computation only executes on one worker.

yuvipanda:
Things from this we have to upstream (from looking at customizations we had to perform with modify_pod_hook):

1. Privileged notebook access (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/300)  (although you probably want FlexVolume or a CSI provider in the long term, see the linked issue)
2. Notebook capabilities add / remove
3. Set default URL for launching (allows us to set JupyterLab instead of notebook as default)
4. Set service account to be used

I'll file bugs in z2jh for these and link from here appropriately.

rabernat:
Should we consider merging this?

jhamman:
Unless there is any opposition, I'll merge this tomorrow afternoon.

PR #83, opened 2018-01-18
Add slurm support WIP

jedwards4b:


rabernat:
Thanks @jedwards4b for getting this started! I really support this work, as I personally use pangeo on two slurm-based clusters (Columbia's Habanero and TACC Wrangler). I currently have a hacky solution (see #79), so I would be happy to replace it with a more robust approach such as what you are developing here.

pangeo/jobqueue.py
```diff
@@ -0,0 +1,126 @@
...
+import toolz
+
+from distributed.utils import tmpfile, ignoring
+
+logger = logging.getLogger(__name__)
+
+class JobQueue(object):
+    """ Base class to launch Dask to a job queue, this class should not be used directly,
+    use inherited class appropriate for your queueing system eg PBScluster or SLURMcluster
+    """
```
rabernat: This seems like the right approach to me: a generic base class with subclasses for the specific scheduler details. 👍

pangeo/jobqueue.py
```diff
@@ -0,0 +1,126 @@
+from contextlib import contextmanager
+import logging
+import subprocess
+import toolz
+
+from distributed.utils import tmpfile, ignoring
+
+logger = logging.getLogger(__name__)
+
+class JobQueue(object):
```
jhamman: pep8: two blank lines before classes

pangeo/jobqueue.py
jhamman: Suggest cleaning up this doc string according to [pep8/pep257]:

```Python
    """Baseclass for launching Dask Distributed Clusters for job queues

    This class should not be used directly, use inherited class appropriate for
    your queueing system (e.g. PBScluster or SLURMcluster).
    """
```

pangeo/jobqueue.py
```diff
@@ -0,0 +1,126 @@
...
+    """ Base class to launch Dask to a job queue, this class should not be used directly,
+    use inherited class appropriate for your queueing system eg PBScluster or SLURMcluster
+    """
+    def __init__(self):
+        logger.error("JobQueue class should not be called directly")
```
jhamman: maybe just raise a `NotImplementedError` here. Or just remove this method.

pangeo/jobqueue.py
```diff
@@ -0,0 +1,126 @@
...
+    def job_file(self):
+        """ Write job submission script to temporary file """
+        with tmpfile(extension='sh') as fn:
+            with open(fn, 'w') as f:
+                f.write(self.job_script())
+            yield fn
+
+    def start_workers(self, n=1):
+        """ Start workers and point them to our local scheduler """
+        outs = []
```
jhamman: `outs` is never used.

pangeo/jobqueue.py
```diff
@@ -0,0 +1,126 @@
...
+        self._call([self._cancelcmd] + list(jobs))
+        for w in workers:
+            with ignoring(KeyError):
+                del self.jobs[w]
+
+    def scale_up(self, n, **kwargs):
+        return self.start_workers(n - len(self.jobs))
+
+    def scale_down(self, workers):
+        if isinstance(workers, dict): # https://github.com/dask/distributed/pull/1659
```
jhamman: I don't think this comment is useful is needed anymore. This feature is not in distributed@master.

pangeo/slurm.py
```diff
@@ -0,0 +1,117 @@
...
+                 extra='',
+                 **kwargs):
+        """ Initialize a SLURM Cluster
+
+        Parameters
+        ----------
+        name : str
+            Name of worker jobs. Passed to `#SBATCH -J` option.
+        queue : str
+            Destination queue for each worker job. Passed to `#SBATCH -p` option.
```
jhamman: pep8: wrap this line at 79 characters (or before).

pangeo/tests/test_slurm.py
```diff
@@ -0,0 +1,68 @@
+import os
+from time import time, sleep
+
+import pytest
+
+from dask.distributed import Client
+from distributed.deploy.adaptive import Adaptive
```
jhamman: Adaptive is not used

pangeo/tests/test_slurm.py
```diff
@@ -0,0 +1,68 @@
...
+    with SLURMCluster(walltime='00:02:00', loop=loop) as cluster:
+        cluster.adapt()
+        with Client(cluster) as client:
+            future = client.submit(lambda x: x + 1, 10)
+            assert future.result(60) == 11
+
+            assert cluster.jobs
+
+            start = time()
+            while len(client.scheduler_info()['workers']) != cluster.config['processes']:
```
jhamman: line length

pangeo/tests/test_slurm.py
```diff
@@ -0,0 +1,68 @@
...
+            start = time()
+            while cluster.jobs:
+                sleep(0.100)
+                assert time() < start + 10
+
+
+@pytest.mark.skipif('PBS_ACCOUNT' in os.environ, reason='PBS_ACCOUNT defined')
+def test_errors(loop):
+    with pytest.raises(ValueError) as info:
+        PBSCluster()
```
jhamman: Should this be `SLURMCluster`?

mrocklin:
Instead of raising many small style comments I recommend that we switch to running flake8 in the tests

mrocklin:
In principle this approach seems good to me.  I'm glad to see the amount to which the two solutions are the same.

jedwards4b:
@jhamman thanks for the review, I think I've addressed all of your issues.

jhamman:
@jedwards4b - this looks good. Thanks for putting this together, I'm excited to try it out.

General question for those interested in this sort of thing. Do we have any way to test these in a CI system? I'm guessing the answer is probably no for running PBS or Slurm but it would be good to have some tests that we can run apart from HPC machines.

Issue #87, opened 2018-01-24
Allow setting a list of images to spawn from when users log in


yuvipanda:
We'd like to:

1. Have a list of docker images specified in our config.yaml
2. Allow users to pick one of them when they launch.

I think the way to do this is:

1. Use http://zero-to-jupyterhub.readthedocs.io/en/latest/advanced.html#arbitrary-code-in-jupyterhub-config-py set extraConfigMap to have a list of docker images in config.yaml
2. Use `extraConfig` (from same link as 1) to make an [options form](http://jupyterhub.readthedocs.io/en/latest/reference/spawners.html?highlight=options#spawner-options-form). See https://github.com/google/kubeflow/blob/master/kubeflow/core/jupyterhub.libsonnet#L19 for an example of the Python code. You're subclassing KubeSpawner and adding more methods.

mrocklin:
This would probably help people who have significantly different stacks than what we currently deploy.  In particular @rsignell-usgs and @jacobtomlinson and @niallrobinson

yuvipanda:
This can happen easily once https://github.com/jupyterhub/kubespawner/pull/137 is merged, so I suggest we wait for it.

jhamman:
Has anyone tried this yet? I suspect this is something that is going to interest a number of us.

jacobtomlinson:
I'm looking at this again during the Pangeo developer meeting sprints. Help would be appreciated!

jacobtomlinson:
It looks like we are waiting on `z2jh` 0.7 for this to make it's way upstream. However the functionality should be in `KubeSpawner` already so we may be able to add it manually for now.

yuvipanda:
@jacobtomlinson 0.7 has been released a few weeks ago!

rabernat:
We should also add options to select different resource options (CPU / GPU / memory).

delgadom:
Could this feature be used to select preemptable/nonpreemptable nodes, or other configurable hardware options?

rabernat:
profile_list! This is the technical solution

https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html?highlight=profile_list

mrocklin:
I think that @rabernat's immediate need isn't so much with different
images, as it is with different resource configurations like high-memory or
GPU nodes.

rabernat:
[...]

PR #95, opened 2018-01-30
Jupyterhub/Dask-Kubernetes setup on AWS

amanda-tan:
Using the setup for GCE but modified to run on AWS using the Heptio Quickstart Template.

### DIFF
--- a/aws/jupyter-config.yaml
+++ b/aws/jupyter-config.yaml
@@ -1,0 +1,54 @@
+proxy:
+  secretToken: "e1e761517a8f4c025144e53be60ffc482e6321e26d8a184c2bab289948b8ba13"
+
+rbac:
+  enabled: false
+
+singleuser:
+  image:
+    name: daskdev/pangeo-notebook
+    tag: latest
+  cmd: ['start-singleuser.sh']
+  extraEnv:
+    EXTRA_PIP_PACKAGES: >-
+
+auth:
+  type: github
+  github:
+    clientId: "1fb1c4f13366624db097"
+    clientSecret: "SECRET"
+    callbackUrl: "https://pangeo-aws.cloudmaven.org/hub/oauth_callback"
+  admin:
+    access: true
+    users:
+      - mrocklin
+      - jhamman
+      - rabernat
+      - yuvipanda
+      - choldgraf
+      - robfatland
+      - aaarendt
+      - atlehr
+      - lsetiawan
+
+hub:
+  extraConfig: |
+    c.KubeSpawner.singleuser_service_account = 'default'
+    c.Spawner.default_url = '/lab'
+    from kubernetes import client
+    def modify_pod_hook(spawner, pod):
+        pod.spec.containers[0].security_context = client.V1SecurityContext(
+            privileged=True,
+            capabilities=client.V1Capabilities(
+                add=['SYS_ADMIN']
+            )
+        )
+        return pod
+    c.KubeSpawner.modify_pod_hook = modify_pod_hook
+
+cull:
+  enabled: true
+  users: false
+  timeout: 1200
+  every: 120
+
--- a/aws/jupyter-config.yaml
+++ b/aws/jupyter-config.yaml
@@ -1,5 +1,5 @@
 proxy:
-  secretToken: "e1e761517a8f4c025144e53be60ffc482e6321e26d8a184c2bab289948b8ba13"
+  secretToken: "SECRET"
 
 rbac:
   enabled: false
--- a/aws/notebook/.daskernetes.yaml
+++ b/aws/notebook/.daskernetes.yaml
@@ -1,0 +1,27 @@
+worker:
+  spec:
+    containers:
+    - args: [dask-worker, --nthreads, '2', --no-bokeh, --memory-limit, 8GB, --death-timeout, '60']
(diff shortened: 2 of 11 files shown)


jhamman:
Thanks @atlehr. 

@rsignell-usgs and @jreadey - looking for thoughts / comments on this.

Nice work to all involved.

cc @yuvipanda and @mrocklin

aws/notebook/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    zict \
+    && conda clean -tipsy
+
+RUN jupyter labextension install @jupyter-widgets/jupyterlab-manager
+
+RUN pip install fusepy click jedi kubernetes --upgrade --no-cache-dir
+
+RUN pip install git+https://github.com/mrocklin/daskernetes@deploy \
+                git+https://github.com/zarr-developers/zarr \
+                git+https://github.com/jhamman/xarray@fix/zarr_set_attrs \
```
jhamman: No need to change immediately but this can be pydata/xarray@master now.

aws/notebook/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+                git+https://github.com/martindurant/gcsfs.git@fuse \
+                git+https://github.com/dask/distributed@master \
+                git+https://github.com/dask/dask@master \
+                --no-cache-dir
```
jhamman: More generally, we can probably move to full releases / conda installs of a bunch of these.

aws/jupyter-config.yaml
```diff
@@ -0,0 +1,54 @@
+proxy:
+  secretToken: "e1e761517a8f4c025144e53be60ffc482e6321e26d8a184c2bab289948b8ba13"
```
jreadey: Not a good idea to have secret checked in!

aws/notebook/.daskernetes.yaml
```diff
@@ -0,0 +1,27 @@
...
+        capabilities:
+          add: [SYS_ADMIN]
+        privileged: true
+      volume_mounts:
+        - mountPath: /dev/fuse
+          name: fuse
+      env:
+        - name: EXTRA_PIP_PACKAGES
+          value: git+https://github.com/dask/distributed
+        - name: GCSFUSE_BUCKET
```
jreadey: Does this translate to S3 bucket?  Use a more generic name for this environment variable?

aws/jupyter-config.yaml
atlehr: Yep, should've checked before committing.

aws/notebook/Dockerfile
rabernat: Is it possible to separate the python environment specification into its own file (environment.yml)? Or does something about docker require the whole DOCKERFILE to be a single self-contained script. Separating the python environment stuff from the more general setup could make it easier to define new environments down the line...

This is partly related to #87.

aws/notebook/.daskernetes.yaml
martindurant: AWS can read from GCS, albeit less performantly, and I don't suppose there is a desire to duplicate the data. Having said that, there will be a need to transfer some form of credentials, if not on google hardware. 

There are s3/fuse solutions out there that, as far as I know, have not been tried in this context.

jhamman:
Something that may be worth pursing here is to mount `s3://nasanex` via fuse. This would give us access to https://aws.amazon.com/public-datasets/nasa-nex/ and may be a nice, larger set of datasets for working on some of these file access issues.

aws/setup.sh
```diff
@@ -0,0 +1,25 @@
...
+helm init --service-account tiller
+kubectl --namespace=kube-system patch deployment tiller-deploy --type=json --patch='[{"op": "add", "path": "/spec/template/spec/containers/0/command", "value": ["/tiller", "--listen=localhost:44134"]}]'
+
+# Get Helm repositories
+helm repo add jupyterhub https://jupyterhub.github.io/helm-chart/
+helm repo update
+
+# Install JupyterHub and Dask on the cluster
+helm install jupyterhub/jupyterhub --version=v0.6.0-9701a90 --name=jupyter --namespace=pangeo -f jupyter-config.yaml
+helm install dask/dask --name=dask --namespace=pangeo -f dask-config.yaml
```
jhamman: @mrocklin - question about `dask-config.yaml`. In the GCP deployment, you have also referenced this file but it is not included in the repository: 

https://github.com/pangeo-data/pangeo/blob/10dc105bae0083968417a8bec2b293ee1e6b4e2a/gce/setup.sh#L16-L18

Conversely, `jupyter-config.yaml` is included: https://github.com/pangeo-data/pangeo/blob/master/gce/jupyter-config.yaml

Any chance this was omitted from your commit/PR by accident?

mrocklin: This file has been removed.  This line should be as well.

jhamman:
After discussing with @amanda-tan / @robfatland - we've decided to close this PR for the time being. The http://pangeo-aws.cloudmaven.org/ jupyterhub will be down "for maintenance" for the foreseeable future. When bandwidth opens up on their end again, they may redeploy the cluster but the timetable for that is TBD. 

@rsignell-usgs and @jreadey are moving forward with AWS-pangeo developments.

PR #100, opened 2018-02-07
en_US.utf8 for slurm job template

jhamman:
I have been finding that this is necessary on Geyser. @jedwards4b or @davidedelvento may know a better way to make sure these environment variables are set but `click` doesn't seem to like the barebones environment.

### DIFF
--- a/pangeo/slurm.py
+++ b/pangeo/slurm.py
@@ -79,6 +79,10 @@ class SLURMCluster(JobQueue):
 #SBATCH -e %(name)s.err
 #SBATCH -o %(name)s.out
 
+export LANG="en_US.utf8"
+export LANGUAGE="en_US.utf8"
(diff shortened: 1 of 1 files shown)


rabernat:
This is ringing a bell for me. I had to do this too on a different cluster I think.

jhamman:
Right, I've seen this on other clusters before too. Whenever jobs get a completely bare environment, this seems to happen.

davidedelvento:
I don't know a better way of doing this and don't have any other comment on this.

jedwards4b:
Should this go into the pbs template as well?

jhamman:
> Should this go into the pbs template as well?

I don't think it is needed. At least on Cheyenne, the base environment seems to have the necessary tools in place to handle this on the fly.

PR #102, opened 2018-02-07
Fix resources in .daskernetes.yaml file

mrocklin:
cc @rabernat you should make this change in your environment.


rabernat:
Just so I understand...merging this PR will only cause this change to be applied to the `.daskernetes.yaml` file in *newly created* user home directories?

mrocklin:
Correct, but you might consider applying this change to your
.daskernetes.yaml file as well

jhamman:
@mrocklin - is this superseded by #115?

mrocklin:
Yes, thanks for the reminder

PR #105, opened 2018-02-07
pleaides launch scripts

rabernat:
Scripts for launching from NASA's pleaides:
https://www.nas.nasa.gov/hecc/support/kb/

### DIFF
--- a/utilities/pleiades/launch-dask-scheduler.sh
+++ b/utilities/pleiades/launch-dask-scheduler.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+#PBS -N dask-sched
+#PBS -W group_list=g26209
+#PBS -l select=1:ncpus=20:mpiprocs=5:model=ivy
+#PBS -l walltime=7:59:00
+#PBS -j oe
+#PBS -m abe
+
+export OMP_NUM_THREADS=4
+export LANG="en_US.utf8"
+export LANGUAGE="en_US.utf8"
+export LC_ALL="en_US.utf8"
+
+# Connect with
+# >>> from dask.distributed import Client
+# >>> client = Client(scheduler_file='~/scheduler.json')
+
+# Setup Environment
+module purge
+source activate pangeo
+
+SCHEDULER=/nobackup/$USER/dask/scheduler.json
+rm -f $SCHEDULER
+mpirun --np 5 dask-mpi --nthreads 4 \
+    --memory-limit 0.2 \
+    --interface ib0 \
+    --local-directory /nobackup/$USER/dask \
+    --scheduler-file=$SCHEDULER
--- a/utilities/pleiades/launch-dask-worker.sh
+++ b/utilities/pleiades/launch-dask-worker.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+#PBS -N dask-worker
+#PBS -W group_list=g26209
+#PBS -l select=1:ncpus=20:mpiprocs=5:model=ivy
+#PBS -l walltime=7:59:00
+#PBS -j oe
+#PBS -m abe
+
+export OMP_NUM_THREADS=4 
+export LANG="en_US.utf8"
+export LANGUAGE="en_US.utf8"
+export LC_ALL="en_US.utf8"
+
+# Setup Environment
+module purge
+source activate pangeo
+
+# Setup dask worker
+SCHEDULER=/nobackup/$USER/dask/scheduler.json
+mpirun --np 5 dask-mpi --nthreads 4 \
+    --memory-limit 0.2 \
+    --interface ib0 \
+    --no-scheduler --local-directory /nobackup/$USER/dask \
+    --scheduler-file=$SCHEDULER
--- a/utilities/pleiades/launch-dask.sh
+++ b/utilities/pleiades/launch-dask.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -e
+
+export LANG="en_US.utf8"
+export LANGUAGE="en_US.utf8"
+export LC_ALL="en_US.utf8"
+
+echo "Launching dask scheduler"
+s=`qsub launch-dask-scheduler.sh`
+sjob=${s}
+echo ${s}
+
+workers=${1:-4}
+echo "Launching dask workers (${workers})"
+for i in $(seq 1 ${workers}); do
+    qsub launch-dask-worker.sh
+done
+
+qstat ${sjob}
+
+# block until the scheduler job starts
+while true; do
+    status=`qstat ${sjob} | tail -n 1`
+    echo ${status}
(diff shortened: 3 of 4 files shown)


utilities/pleiades/launch-dask-scheduler.sh
```diff
@@ -0,0 +1,28 @@
...
+export LANGUAGE="en_US.utf8"
+export LC_ALL="en_US.utf8"
+
+# Connect with
+# >>> from dask.distributed import Client
+# >>> client = Client(scheduler_file='~/scheduler.json')
+
+# Setup Environment
+module purge
+source activate pangeo
```
mrocklin: Can this be done in the user's environment instead?  It seems odd to bake in the name of a particular conda environment.

utilities/pleiades/launch-dask-scheduler.sh
```diff
@@ -0,0 +1,28 @@
...
+
+SCHEDULER=/nobackup/$USER/dask/scheduler.json
+rm -f $SCHEDULER
+mpirun --np 5 dask-mpi --nthreads 4 \
+    --memory-limit 0.2 \
```
mrocklin: I suspect that you might want to just drop this.  Dask will default to multiplying memory by `nthreads / total-logical-cores`

utilities/pleiades/launch-dask-worker.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+#PBS -N dask-worker
+#PBS -W group_list=g26209
+#PBS -l select=1:ncpus=20:mpiprocs=5:model=ivy
+#PBS -l walltime=7:59:00
+#PBS -j oe
+#PBS -m abe
+
+export OMP_NUM_THREADS=4
```
mrocklin: I generally recommend setting this to 1, otherwise Dask.array can call parallelized operations in parallel again, leading to over-saturation.

utilities/pleiades/launch-dask.sh
```diff
@@ -0,0 +1,40 @@
...
+
+echo "Launching dask scheduler"
+s=`qsub launch-dask-scheduler.sh`
+sjob=${s}
+echo ${s}
+
+workers=${1:-4}
+echo "Launching dask workers (${workers})"
+for i in $(seq 1 ${workers}); do
+    qsub launch-dask-worker.sh
```
mrocklin: You might want to look into array jobs for `qsub`.  I think you might want `qsub -t 1-$workers` here instead of the for loop.

utilities/pleiades/setup-jlab.py
```diff
@@ -0,0 +1,77 @@
...
+    print('Run the following command from your local machine:')
+    print(f'ssh -N -L {jlab_port}:{host}:{jlab_port} '
+          f'-L {dash_port}:{host}:8787 {user}@{hostname}')
+    print('Then open the following URLs:')
+    print(f'\tJupyter lab: http://localhost:{jlab_port}')
+    print(f'\tDask dashboard: http://localhost:{dash_port}', flush=True)
+
+
+if __name__ == '__main__':
+    cli()
```
mrocklin: Is this different from the other setup_jlab script?  If not then is there a way to avoid duplication?

jhamman:
@rabernat - two things.

2) I'm glad you are working on pleaides. I have a project spinning up now that has an allocation there so I was going to do this myself. Glad you are leading the charge.
1) you might try using the `PBSCluster` class from a standard notebook interface. I'm generally finding this is a preferable interface.

rabernat:
FYI, the goal here is to get the clsuter up and running fast and document something minimal about it.

I agree that the launch scripts could be improved / refactored a lot. Not sure I have time to take that on, as I am kind of eager to do "actual science" right now ;)

rabernat:
@jhamman where are we on documentation for `PBSCluster`?

utilities/pleiades/launch-dask-scheduler.sh
rabernat: but how would we guarantee that the user's environment actually has dask?

mrocklin: Presumably this would err if they didn't have dask.

Calling `source activate pangeo` doesn't guarantee this either though, unless they happen to have the same environment as you do.

rabernat:
@mrocklin - I removed the `--memory-limit` flag from the dask mpi scripts. Each worker is now only getting 6 GiB of RAM. There are five workers per node, and each node has 64 GB memory. This doesn't seem right.

mrocklin:
We divide the total memory (as identified by psutil) by the total number of logical cores (as identified by multiprocessing).  Is there something off in the logic below?

https://github.com/dask/distributed/blob/4f61ff588de5f8ee504f6a936396b91e7cfc97f7/distributed/worker.py#L2667-L2668

```python
def parse_memory_limit(memory_limit, ncores):
    if memory_limit is None:
        return None
    if memory_limit == 'auto':
        memory_limit = int(TOTAL_MEMORY * min(1, ncores / _ncores))
    with ignoring(ValueError, TypeError):
        x = float(memory_limit)
        if isinstance(x, float) and x <= 1:
            return int(x * TOTAL_MEMORY)

    if isinstance(memory_limit, (unicode, str)):
        return parse_bytes(memory_limit)
    else:
        return int(memory_limit)
￼```

**Comment by rabernat** [comment]
These are ivybridge processors, with 20 cores:
https://www.nas.nasa.gov/hecc/support/kb/ivy-bridge-processors_445.html

`multiprocessing.cpu_count()` returns 40.

Is this because of hyperthreading? (Also, what is hyperthreading?) Should I really be using twice the number of threads as I am now?

**Comment by jhamman** [comment]
> @jhamman where are we on documentation for PBSCluster

None yet. We have been discussing the PBSCluster and SlurmCluster classes to a standalone `dask-jobqueue` repo. Until then, we're basically in a holding pattern on developing detailed documentation (that could be changed if needed though). Basic usage is below:

```
# launch a jupyter notebook server in whatever way you like/can
# this may or not be on a compute node depending on your system configuration
# do the standard ssh port forwarding thing
```

Then from inside your jupyter notebook:

```Python
from pangeo import PBSCluster
from distributed import Client

cluster = PBSCluster(project='something', ...)
cluster.autoscale()
client =  Client(cluster)

# now proceed like normal
```

mrocklin:
> Is this because of hyperthreading? 

Yes, probably

> Also, what is hyperthreading?

In some workloads these chips will probably be a bit faster if you have twice as many threads as you have physical cores.  This will be something like 25% faster, not 100% faster.

> Should I really be using twice the number of threads as I am now?

Possibly, yes.

jhamman:
@rabernat - do we still want to merge this? Have you tried using dask-jobqueue on pleaides yet?  

More broadly, I wonder if we need to have these utilities in this repo at all?

PR #115, opened 2018-02-12
Update pangeo gce deployment to new versions

mrocklin:
This does a few things:

1.  Updates versions of dask/distributed to latest release
2.  Uses new mainline version of daskernetes, rather than a custom branch
3.  Replaces the .daskernetes.yaml file with a worker-template.yaml file
4.  Handles kubernetes resources properly, which engages autoscaling
5.  Adds vim to the notebook
6.  Adds scikit-image to the environments, including a new notebook

Basic workflows still seem to operate fine.
Any customization that was placed into the .daskernetes.yaml file is no
longer used.


gce/notebook/Dockerfile
```diff
@@ -2,17 +2,16 @@ FROM jupyter/base-notebook
...
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
 USER $NB_USER
 
 RUN conda install --yes -c conda-forge \
     cytoolz \
-    dask \
-    distributed \
+    dask=0.17.0 \
+    distributed=1.21.0 \
```
jhamman: just curious, what's the reason for pinning these versions

jhamman:
@mrocklin - looks good to me.

gce/notebook/Dockerfile
mrocklin: 1. These are newest
2. We should probably get in the habit of pinning more so that we're explicit about what was running when

gce/worker/prepare.sh
```diff
@@ -0,0 +1,27 @@
...
+    echo "no environment.yml"
+fi
+
+if [ "$EXTRA_CONDA_PACKAGES" ]; then
+    echo "EXTRA_CONDA_PACKAGES environment variable found.  Installing."
+    /opt/conda/bin/conda install $EXTRA_CONDA_PACKAGES
+fi
+
+if [ "$EXTRA_PIP_PACKAGES" ]; then
+    echo "EXTRA_PIP_PACKAGES environment variable found.  Installing".
```
jhamman: Is this the recommended way of specifying custom packages for workers now? As apposed to specifying them in the `.daskernetes.yaml`. 

cc @atlehr

mrocklin: You can specify a complete custom docker image, or you can specify environment variables like `EXTRA_PIP_PACKAGES`.  Those environment variables get picked up when the docker container starts and runs this script.  This is how it has always worked.

Note that the `.daskernetes.yaml` file is no longer used.  Instead you point to a file like worker-template.yaml.  This is specifyable when you create a cluster.  See http://daskernetes.readthedocs.io/en/latest/

PR #122, opened 2018-02-20
Add instructions to update the pangeo.pydata.org deployment

mrocklin:


mrocklin:
cc @rabernat

rabernat:
Thanks Matt. It is great to have this dumped out of your brain and into a public place!

Eventually we will want to incorporate something like this into the sphinx site. But for now I am happy to have a page of standalone instructions.

gce/update.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+### Push to dockerhub
+
+You can push both images to docker hub:
+
+```bash
+docker push dsakdev/pangeo-worker:TAG
+docker push dsakdev/pangeo-notebook:TAG
+```
```
rabernat: It sounds like it is possible to update the docker images without ever syncing up the pangeo repo. My worry is that, if multiple people start updating the deployment this way, we will break something and not be able to roll back because we don't have the configuration under version control.

A better scenario would be that the docker files and example notebooks are updated via PR, and the rest of the build / deploy process occurs automatically via some sort of CI magic. I recognize that this may be difficult to set up and therefore not practical right now, but I just wanted to raise the possibility.

If we don't go that route, maybe we can update these instructions to remind whoever is doing this to also push the changes to github.

mrocklin: I agree that this would be a good way to manage things.  I'm happy to help advise others if they take this on.

PR #146, opened 2018-03-07
update notebook dockerfile

### LINKED ISSUE #145: enable bokeh / holoviews / geoviews in pangeo.pydata.org
Next Monday we will be giving a demo to some NASA folks of pangeo.pydata.org. I would like to show some example of visualization with holoviews. But, as pointed out in #144, the bokeh jupyterlab extension evidently doesn't work. And I don't think that holoviews is installed at all.

cc @mrocklin, @jbednar

rabernat:
Fixes #145 

This is what I am going to use to rebuild the notebook docker image. I just wanted to put this up so I can get feedback on how I modified to dockerfile. I'm new to docker, but it seems pretty straightforward.

### DIFF
--- a/gce/notebook/Dockerfile
+++ b/gce/notebook/Dockerfile
@@ -29,7 +29,14 @@ RUN conda install --yes -c conda-forge \
     zict \
     && conda clean -tipsy
 
-RUN jupyter labextension install @jupyter-widgets/jupyterlab-manager @jupyterlab/hub-extension
+# https://github.com/pangeo-data/pangeo/issues/145#issuecomment-371215319
+RUN conda install -c ioam/label/dev holoviews geoviews
+RUN conda install -c bokeh/label/dev bokeh
+
+RUN jupyter labextension install \
+    @jupyter-widgets/jupyterlab-manager \
+    @jupyterlab/hub-extension \
+    @pyviz/jupyterlab_holoviews
 
 RUN pip install fusepy click jedi kubernetes --upgrade --no-cache-dir
 
@@ -41,7 +48,7 @@ RUN pip install daskernetes==0.1.3 \
 
 RUN pip install distributed==1.21.1 \
                 dask==0.17.1 \
-                --no-dependencies \ 
+                --no-dependencies \
                 --no-cache-dir
 
 RUN pip install git+https://github.com/jupyterhub/nbserverproxy --no-cache-dir


gce/notebook/Dockerfile
```diff
@@ -29,7 +29,14 @@ RUN conda install --yes -c conda-forge \
     zict \
     && conda clean -tipsy
 
-RUN jupyter labextension install @jupyter-widgets/jupyterlab-manager @jupyterlab/hub-extension
+# https://github.com/pangeo-data/pangeo/issues/145#issuecomment-371215319
+RUN conda install -c ioam/label/dev holoviews geoviews
```
philippjfr: Just to be totally clear, up to date HoloViews and GeoViews releases do not yet exist. I'll ping you as soon as they are available.

mrocklin: To keep the size of the image small you'll want to include the `&& conda clean -tipsy` bit from above.  You'll also want to verify that this doesn't bring down a lot of new packages and replace those above. 

If possible, I might suggest using `pip` (if holoviews is easy to install) or including this in the conda install line above.

mrocklin:
You might also want to test things locally.  I think that @jhamman has created a how-to guide with minikube.

gce/notebook/Dockerfile
philippjfr: HoloViews should be easy to install but we haven't yet made any plans to make ``--pre`` releases on PyPI, certainly an option though. The real worry will be GeoViews which a) is not currently available using pip and b) has dependencies on the entire geostack (including GDAL, cartopy etc.), which is pretty heavy-weight.

mrocklin: Well then we probably shouldn't install geoviews here if we can avoid it.  @rabernat are your needs satisfied with just holoviews?

rabernat:
Testing locally seems like considerably more work than just rebuilding and pushing the one docker image. Is this recommended as a way to avoid accidentally screwing things up?

Alternatively / in addition to local testing, we could consider having separate dev / production environments in the cloud.

rabernat:
There is no need to have holoviews stuff on the worker image, correct?

gce/notebook/Dockerfile
rabernat: I am ok with just holoviews for now.

What if we just pip install holoviews from github master. Will this work?

philippjfr: Actually after double checking it turns out I was wrong about GDAL, it only depends on ``geos`` and ``proj4`` which aren't too large.

mrocklin:
Not that I can tell, no.

gce/notebook/Dockerfile
rabernat: @philippjfr have you made releases yet? Or do you recommend I install from pip?

rabernat:
Do we want to incorporate some of the environment modifications discussed in #148 into this PR?

gce/notebook/Dockerfile
philippjfr: Okay I can't say when that will be done as I'm not the one working on it, so maybe use pip after all.

philippjfr: HoloViews 1.10.0a1 is now up on the ``ioam/label/dev`` channel.

rabernat:
Closed in favor of #149

PR #158, opened 2018-03-14
Add holoviews

mrocklin:


mrocklin:
I pushed this yesterday but failed to submit a PR.

gce/notebook/Dockerfile
```diff
@@ -33,11 +36,11 @@ RUN conda install --yes \
 RUN pip install fusepy click jedi kubernetes --upgrade --no-cache-dir
 
 RUN pip install daskernetes==0.1.3 \
-                distributed==1.21.3 \
+                git+https://github.com/dask/distributed@9ade11e41b5c63a5977230a01ad5b2ef15517238 \
                 git+https://github.com/dask/dask-kubernetes@5ba08f714ef38e585e9f2038b6be530c578b96dd \
                 git+https://github.com/zarr-developers/zarr \
                 git+https://github.com/pydata/xarray \
-                git+https://github.com/ioam/holoviews \
+                git+https://github.com/ioam/holoviews@range_optimizations \
```
philippjfr: The PR has now been merged and the branch deleted, hadn't considered you'd be using it so I'll restore it for now.

rabernat: should we just be using the master branch then?

philippjfr: Yes, master should be good now. I also noticed that there are still some issues with the holoviews widgets in JupyterLab on pangeo. Another user was reporting the same thing, but I wasn't reproduce it locally, so that should help me fix it at least.

rabernat: @mrocklin would you like to update this? I would like to add a notebook to the example gallery today. I will have to rebuild the docker image, so I want to make sure I'm using the most up-to-date dockerfile.

PR #163, opened 2018-03-16
easier customization of conda env/module

jbusecke:
This PR implements: 

1) A single script to load necessary modules and conda environments (which gets sourced in all other scripts). This makes it much easier to customize these variables (I run a local conda install with a different name)

2) Improve compatibility issue with python < 3.6  in setup_jlab.py (f string formatting produces an error)

### DIFF
--- a/utilities/habanero/mod_env_setup.sh
+++ b/utilities/habanero/mod_env_setup.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-load module anaconda
+module load anaconda
 source activate pangeo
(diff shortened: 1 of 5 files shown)


utilities/habanero/mod_env_setup.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+load module anaconda
```
rabernat: isn't it `module load anaconda`?

jbusecke: You are right. Let me fix that real quick.

mrocklin:
What is the status of this?

rabernat:
This collection of cluster-specific scripts is messy but useful to us. I'm not sure if they belong in the main pangeo repo long term, and there are no tests. But I will merge for now.

Issue #131, opened 2018-02-23
automate building of docker images when dockerfiles are updated


rabernat:
One way to begin to automate the deployment of our GCE would be to automatically build the docker images whenever the dockerfiles are updated in this repo. It looks like that is not too hard to do:
https://docs.docker.com/docker-hub/builds/

That would remove some of the manual steps in the [cluster update process](https://github.com/pangeo-data/pangeo/blob/master/gce/update.md).

xref #128, #117

mrocklin:
We might also want to run `helm upgrade` if we're able to securely store secrets in a CI server somewhere.  

We might also want to move this to a different repository that is specific to the pydata.pangeo.org deployment and keep this repository suitably general.

rabernat:
Can you clarify what you mean by "run `helm upgrade`"? Run where? Is it possible to automate that as well somehow?

My understanding is that `helm upgrade` had to be run manually / locally. I don't know enough CI to know how to automate that step.

PR #172, opened 2018-03-21
Add basic RBAC

tjcrone:
As per suggestions by @jacobtomlinson, add basic RBAC setup.

### DIFF
--- a/gce/jupyter-config.yaml
+++ b/gce/jupyter-config.yaml
@@ -24,7 +24,7 @@ singleuser:
     enabled: true
 
 rbac:
-  enabled: false
+  enabled: true
 
 hub:
   extraConfig: |
--- a/gce/jupyter-config.yaml
+++ b/gce/jupyter-config.yaml
@@ -28,7 +28,7 @@ rbac:
 
 hub:
   extraConfig: |
-    c.KubeSpawner.singleuser_service_account = 'default'
+    c.KubeSpawner.singleuser_service_account = 'daskkubernetes'
     c.Spawner.default_url = '/lab'
     from kubernetes import client
     def modify_pod_hook(spawner, pod):
--- a/gce/jupyter-config.yaml
+++ b/gce/jupyter-config.yaml
@@ -23,9 +23,6 @@ singleuser:
   cloudMetadata:
     enabled: true
 
-rbac:
-  enabled: true
-
 hub:
   extraConfig: |
     c.KubeSpawner.singleuser_service_account = 'daskkubernetes'
--- a/gce/setup.sh
+++ b/gce/setup.sh
@@ -17,8 +17,8 @@ helm repo update
 # Install JupyterHub and Dask on the cluster
 helm install jupyterhub/jupyterhub --version=v0.6.0-9701a90 --name=jupyter --namespace=pangeo -f jupyter-config.yaml -f secret-config.yaml
 
-# create the kubernetesdask service account and role bindings
-echo "Installing service account for dask-kubernetes."
+# create the daskkubernetes service account and role bindings
+echo "Installing service account for daskkubernetes."
 kubectl create -f dask-kubernetes-serviceaccount.yaml
 
 # Look for publised services.  Route domain name A records to these IPs.
--- a/gce/setup.sh
+++ b/gce/setup.sh
@@ -1,5 +1,5 @@
 # Start cluster on Google cloud
-gcloud container clusters create pangeo --num-nodes=10 --machine-type=n1-standard-2 --zone=us-central1-b  --cluster-version=1.9.3-gke.0
+gcloud container clusters create pangeo --no-enable-legacy-authorization --num-nodes=10 --machine-type=n1-standard-2 --zone=us-central1-b  --cluster-version=1.9.3-gke.0
 
 gcloud container clusters get-credentials pangeo --zone us-central1-b --project pangeo-181919
 
@@ -16,6 +16,10 @@ helm repo update
 
 # Install JupyterHub and Dask on the cluster
 helm install jupyterhub/jupyterhub --version=v0.6.0-9701a90 --name=jupyter --namespace=pangeo -f jupyter-config.yaml -f secret-config.yaml
+
+# create the kubernetesdask service account and role bindings
+echo "Installing service account for dask-kubernetes."
+kubectl create -f dask-kubernetes-serviceaccount.yaml
 
 # Look for publised services.  Route domain name A records to these IPs.
 kubectl get services --namespace pangeo
--- a/gce/dask-kubernetes-serviceaccount.yaml
+++ b/gce/dask-kubernetes-serviceaccount.yaml
@@ -1,0 +1,32 @@
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: daskkubernetes
+  namespace: pangeo
+
+---
+
+kind: Role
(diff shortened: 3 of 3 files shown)


mrocklin:
Do we need to change the rbac entry in the jupyter-config.yaml file as well?

tjcrone:
Yes, thanks for catching that. I'll fix.

tjcrone:
If you would like me to squash into a single commit, I can do that. Please let me know. Would also be good to have @jacobtomlinson review if he has time.

mrocklin:
We can always squash with the green button.  No reason to do so manually.

rabernat:
🙌

gce/jupyter-config.yaml
```diff
@@ -24,11 +24,11 @@ singleuser:
     enabled: true
 
 rbac:
-  enabled: false
+  enabled: true
 
 hub:
   extraConfig: |
-    c.KubeSpawner.singleuser_service_account = 'default'
+    c.KubeSpawner.singleuser_service_account = 'daskkubernetes'
```
jacobtomlinson: You could move this into the helm config itself instead of the `extraConfig` code as this config option is supported in `z2jh`.

```
singleuser:
    serviceAccountName: daskkubernetes
```

gce/setup.sh
```diff
@@ -1,5 +1,5 @@
 # Start cluster on Google cloud
-gcloud container clusters create pangeo --num-nodes=10 --machine-type=n1-standard-2 --zone=us-central1-b  --cluster-version=1.9.3-gke.0
+gcloud container clusters create pangeo --no-enable-legacy-authorization --num-nodes=10 --machine-type=n1-standard-2 --zone=us-central1-b  --cluster-version=1.9.3-gke.0
```
jacobtomlinson: Yay!

gce/setup.sh
```diff
@@ -17,6 +17,10 @@ helm repo update
 # Install JupyterHub and Dask on the cluster
 helm install jupyterhub/jupyterhub --version=v0.6.0-9701a90 --name=jupyter --namespace=pangeo -f jupyter-config.yaml -f secret-config.yaml
 
+# create the daskkubernetes service account and role bindings
+echo "Installing service account for daskkubernetes."
+kubectl create -f dask-kubernetes-serviceaccount.yaml
```
jacobtomlinson: TL;DR This is an anti-pattern, but probably fine for now.

I'm slightly uncomfortable having additional manifests being added manually as you no longer simply run `helm delete <release> --purge` to remove pangeo. It would be better to have the whole deployment contained in a single helm installation. This could be a good opportunity to create a `pangeo` helm chart.

We currently have a `jadepangeo` [helm chart](https://github.com/met-office-lab/jade-pangeo) which has `jupyterhub/jupyterhub` as a dependency and adds some extra manifests. I would like to get to a position where there is an offical chart which depends on jupyterhub and then we can make `jadepangeo` depend on `pangeo/pangeo`. However this is beyond the scope of this PR so I'll raise a new issue for start work on a PR.

gce/jupyter-config.yaml
```diff
@@ -24,11 +24,11 @@ singleuser:
     enabled: true
 
 rbac:
-  enabled: false
+  enabled: true
```
jacobtomlinson: This defaults to true so you could omit this.

gce/jupyter-config.yaml
jacobtomlinson: Perhaps. This is working on our stack.

To be honest I'm not worried either way.

jacobtomlinson: Just a thought did you remove the ` c.KubeSpawner.singleuser_service_account = 'default'` line when setting within the `singleuser` section? Things in the extra config will override the helm config values.

tjcrone:
@jacobtomlinson, one question I have is, does the role as currently defined limit the pod operations only to the pods that the user owns, or can these operations (verbs) be applied to any pod in the cluster?

gce/setup.sh
jacobtomlinson: See #178

gce/jupyter-config.yaml
tjcrone: Pretty sure I did, but I can try again.

jacobtomlinson:
They can be applied to any pod in the namespace. Therefore it might be sensible to have a separate namespace for dask workers.

tjcrone:
Okay. A separate namespace for workers sounds like it would be a significant increase in security. Another question I have is, what are the steps to running containers as unprivileged?

jacobtomlinson:
Well it looks like the only requirement for privileged containers currently is to allow users to mount FUSE filesystems within their containers.

We are using custom [FUSE flex volume drivers](https://github.com/met-office-lab/s3-fuse-flex-volume#usage-examples) which means that S3 is mounted onto the host instead and then volumed into an unprivileged container. So we could need to create equivalent drivers for the Google platform.

tjcrone:
Is it a crazy idea to launch every notebook pod into its own unique namespace, where it has control over the pods in that namespace but zero control over any other pods in any other namespace?

jacobtomlinson:
Not crazy at all. I have considered this before.

It will require changes to the [`KubeSpawner`](https://github.com/jupyterhub/kubespawner) as it will need to be able to create and remove namespaces too. It also makes assumptions about how the Kubernetes cluster is being managed. For example some people may have been given access to a single namespace on a shared cluster in their org, but others may have dedicated clusters like those used in this project.

Being able to delete a namespace is a pretty big deal as it deletes all resources within it, so giving the `KubeSpawner` that power is a big deal.

mrocklin:
> It will require changes to the KubeSpawner as it will need to be able to create and remove namespaces too

cc @yuvipanda

jacobtomlinson:
This PR is definitely a step in the right direction. Even more can be done going forwards.

I'm keen for it to be merged ASAP so I can incorporate it into my helm PR.

mrocklin:
@jacobtomlinson both you and @tjcrone should have merge privileges.  I encourage you both to use them liberally.  This system is still experimental.  I think it's more imprtant to move quickly than to keep things from breaking, especially in this direction.

If folks are waiting on me to merge things like this we'll probably end up waiting a while.  I'll be especially busy in the coming month, and so am keen to ensure that others feel empowered to make large changes here.

jacobtomlinson:
@mrocklin I'm happy to merge but I am currently unable to test on GCE or update the demo platform.

I'm not sure how you want to handle that? Perhaps it would be good to set up Travis CI or some other CI/CD to run tests and keep the platform up to date?

tjcrone:
Roger that @mrocklin. This PR has had enough discussion and review that I am comfortable merging. Cheers.

mrocklin:
@jacobtomlinson I have just given jacob.tomlinson@informaticslab.co.uk edit rights on the project.

rabernat:
This PR has been merged. But has the cluster been re-deployed with the new settings?

jacobtomlinson:
Not as far as I'm aware.

rabernat:
@jacobtomlinson how do you manage deployment at the met office? Do you manually update your cluster every time you change the configuration, or do you have some automated way to do this? (see #131)

I worry that we are making updates to this configuration without actually testing it live.

jacobtomlinson:
I manually do upgrades using helm. As @mrocklin suggests in #131 this could be automated in CI/CD.

PR #186, opened 2018-03-28
bump gcsfs version slightly

mrocklin:


rabernat:
This is old but I presume it should still be merged, correct?

PR #208, opened 2018-04-11
Added ML example notebook

TomAugspurger:
Adding an example using dask-ml.

cc @mrocklin

### DIFF
--- a/gce/notebook/Dockerfile
+++ b/gce/notebook/Dockerfile
@@ -12,6 +12,7 @@ RUN conda install --yes -c ioam -c bokeh/channel/dev \
     cytoolz \
     datashader \
     dask=0.17.2 \
+    dask-ml \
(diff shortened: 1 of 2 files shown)


rabernat:
It's great to see an ML example here, and I'm happy to merge this.

In the long term, it would be even better if we could get a dask-ml example that uses environmental data!

mrocklin:
This notebook should be modified to use `dask_kubernetes.KubeCluster`

> In the long term, it would be even better if we could get a dask-ml example that uses environmental data!

I hope so too.  Hopefully having an example with basic ML work encourages others to pick things up and get us closer to that goal.

TomAugspurger:
Updated the cluster.

I'd love to use environmental data. Happy to hear suggestions about datasets / topics, but I personally won't have time to write up a notebook using them for a while.

PR #214, opened 2018-04-13
Use a weighted mean for the sealevel example

willirath:
With a lot of the signal from high latitudes, not accounting for the cosine of the latitude will overestimate the increase:

sla_diff

Weighting adds some complexity to the example.  So I'm not sure if this PR really helps with the notebook being a nice and instructive demo...  On the other hand it nicely shows that Dask also handles arithmetics beyond the more obvious reductions.

### DIFF
--- a/gce/notebook/examples/sea-surface-height.ipynb
+++ b/gce/notebook/examples/sea-surface-height.ipynb
@@ -46,7 +46,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "gcsmap = gcsfs.mapping.GCSMap('pangeo-data/dataset-duacs-rep-global-merged-allsat-phy-l4-v3-alt')\n",
+    "gcsmap = gcsfs.mapping.GCSMap(\n",
+    "    'pangeo-data/dataset-duacs-rep-global-merged-allsat-phy-l4-v3-alt')\n",
     "ds = xr.open_zarr(gcsmap)\n",
     "ds"
    ]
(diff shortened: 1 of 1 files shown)


rabernat:
😄 Top marks for keeping us honest!

I debated whether to perform weighting in this example. As you say, it is a balance between scientific accuracy vs. complexity.

However, I think there is a mistake in your weighting function. The line
```python
coslat = xr.ufuncs.cos(xr.ufuncs.deg2rad(lat)).where(da)
```
is not masked. The `.where(da)` does nothing. To make it work as you desire, it needs to be
```python
coslat = xr.ufuncs.cos(xr.ufuncs.deg2rad(lat)).where(~da.isnull())
```

You can see the consequences of this error by taking the weighted mean of an array of ones:
```python
ones = xr.ones_like(ds.sla[0]).where(~ds.sla[0].isnull())
_coslat_weighted_mean(ones).load()
```
We can all agree that the mean of a uniform field of ones, weighted or unweighted, should be one. But, for your original function, this gives 0.69. If you modify it as I suggested, you get 1 exactly.

What are the consequences for the sea level curve?
image

mrocklin:
It's great seeing the same old nuanced discussions occur but now on much
larger datasets :)  I feel like we've achieved a milestone where we've
stopped talking about performance, scale, data formats etc. and are instead
talking about the results of various nuanced analyses and their impacts on
interpretations.

rabernat:
The difference in weighting amounts two cm of sea level! That could be the difference between flooding the subway or not during the next NYC hurricane!

mrocklin:
Nuance!

willirath:
I should have become sceptical at this large a gap... :)

mrocklin:
What is the status of this?

fmaussion:
I think it should still be weighted - for educational purposes at least

rabernat:
Let's have both. First we show the "naive" average and then the weighted average.

@willirath, could you update the notebook.

willirath:
Yes, I'll do that.


Am 7. Mai 2018 17:00:47 schrieb Ryan Abernathey <notifications@github.com>:

jhamman:
@mrocklin - can this be closed following #292?

rabernat:
I don't think this can be closed. #292 is a use-case notebook for the sphinx documentation site. This is a demo notebook for the docker image. Unfortunately, these two repositories of notebooks are not (yet) integrated.

Issue #39, opened 2017-11-15
we need a data catalog


rabernat:
At this point we have multiple datasets on multiple different systems.

It would be great to have some kind of catalog interface to avoid having to manually change the path names every time. For example.

```python
from pangeo import catalog
ds = xr.open_mfdataset(**catalog.dataset('CM2.6'))
```

mrocklin:
Perhaps just some yaml file on the network file system that we refer against as well as some tiny library that reads that file?

Also cc'ing @seibert , who thinks about these things within Anaconda Inc.  @seibert these tend to be glob-string paths pointing to NetCDF files.  All machines have uniform access to them, so there is no need for remote data access.  Modeling data types would be difficult due to the complexity of the underlying files.

mmccarty:
I had a quick email exchange with @jhamman, who pointed me to this issue. Anaconda has a new open source project to provide some functionality for building data catalogs [Intake](http://intake.readthedocs.io/en/latest/overview.html). 

This week, I have been investigating Intake for another project and started a plugin for [NetCDF](https://github.com/ContinuumIO/intake-netcdf). So far, `intake-netcdf` is just a thin wrapper around `xarray` since most of the file I/O and dask functionality in Intake is already provided with `xarray`. This leaves the advantage of the approach to be limited to the cataloging aspects. I'm wondering if this community has any alternatives or input on this approach?

rabernat:
Sounds very promising @mmccarty! The cataloging features of Intake could definitely solve some of the problems we are contemplating.

One complication for Pangeo is that most of our cloud-based data is actually stored in zarr format, not netCDF. So my recommendation would be to broaden the scope beyond netCDF to include any format that can be read by xarray. In the docs you pointed to, it says
> Intake supports loading data into standard Python containers, depending on the data source:
> - Pandas Dataframes - tabular data
> - NumPy Arrays - tensor data
> - Python lists of dictionaries - semi-structured data
> - (possibly more in the future)

There is a strong analogy between pandas and xarray. Pandas.Dataframe is the fundamental python data structure. The pandas data itself can come from csv, hdf, xls, many different file formats. Likewise, xarray data can come from netcdf, hdf, geotiff, zarr, etc. What would be great is to have an xarray data catalog that is agnostic to the underlying file format.

We have also discussed Quilt in #162, which I guess aims to solve a similar problem

countdorkula:
[...]

PR #231, opened 2018-05-01
Example of using intake in a notebook

martindurant:
Ref : #39 

Here is an example of how this would work. I have only edited the `xarray-data` notebook. Essentially, all of the options that were in the original open function calls go into arguments in the catalogue file. We use the same conventions for URLs as dask main, and (read-only) mappers are created automatically. This latter step is probably something that should eventually be made generic for all file-systems.


jhamman:
@martindurant - thanks for putting this together. A few thoughts for you and @rabernat and @mrocklin: 

1. I think the abstraction away from the gcsfs / open_zarr piece is really nice here. Do we, however, run the risk of sweeping some important details under the rug? The process of connecting to these data sources is quite new and it may be instructive for some users to see both the intake and gcsfs/brute-forge methods side by side.
1. There are edits to the docker files here. This will require redeploying the cluster for these changes to take. Do we have any hesitations adding intake as a dependency? (I don't).
1. How do we make sure everyone get's access to this yaml file. Can we host this somewhere on the web? I guess right now we're copying this over through the docker image but it may be better (long term) to have it stored elsewhere.

mrocklin:
> There are edits to the docker files here. This will require redeploying the cluster for these changes to take. Do we have any hesitations adding intake as a dependency? (I don't).

Short term presumably people can `pip install intake` in their local environments to try things out?

mrocklin:
Additionally, we might want to control that for the user as an environment variable

    export INTAKE_CATALOG=/path/to/default/catalog

This way they just have to do something like the following:

```python
import intake
catalog = intake.Catalog()
catalog.my_dataset
```

Honestly, it wouldn't be unwelcome to skip the middle line

```python
import intake  # looks for environment variable, auto-populates global catalog
intake.my_dataset
```

That might be too much though...

martindurant:
We do have the concept of a built-in catalogue and of conda packages that populate it, so you can conda-install and have the entries available under `intake.cat`; these packages may include data or download their own data or point to remote data. That would be another way to distribute a pangeo cat.

martindurant:
Of course, intake is just reaching release now, and we are working on the documentation to make these points clear.

mrocklin:
Can we populate the built-in catalog from the remote web resource?  My
guess is that it will be easier for us to publish a yaml file on some
github repository than it will be for us to publish conda packages

martindurant:
https://github.com/ContinuumIO/intake/pull/87

jacobtomlinson:
@niallrobinson

gce/notebook/examples/sea-surface-height.ipynb
```diff
@@ -51,6 +51,19 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "### Alternate method to load the data using Intake\n",
+    "### Where the remote YAML spec points to the data on GCS\n",
+    "# import intake\n",
+    "# ds = intake.Catalog('gcs://mdtemp/pangeo.yaml').sea_surface.to_dask()\n",
```
mrocklin: Thanks for doing this.

Thoughts on using a file in this github repository instead?  Perhaps in `pangeo/gce/catalog.yaml`?  I suspect that more people within the community are familiar with editing github repositories than modifying files on GCS.

martindurant: That would work fine as well. I don't know where the right place for this would be, so happy to use whatever is convenient.

martindurant:
For the time being, the catalog file is checked in here too, so it can just be a local file in the notebook or point to the remote github location, which would then update whenever the repo was updated.

martindurant:
Hah, it turns out that githubusercontent supplies incorrect HTTP header information: the Content-Length is short, and this truncates the catalogue text. I'll put a fix into dask, since that information should not be required in a case like this, where reading the whole file start-to-end.

rabernat:
Are we ready to merge this? Has someone tried building the docker image with this changes and confirmed it works?

martindurant:
At the moment, there are no instructions to install intake/intake-xarray - I am not sure where they would go. Also, the URL is still directly to GCS, not to github as mrocklin suggested, but of course the catalogue isn't in the master branch yet, so that would not work until after merging.

mrocklin:
My understanding is that this would work if we do the following:

1.  Install intake in docker image
2.  Update dask to master in docker image
3.  Change links to point to the github web address

rabernat:
Thanks for clearing that up @mrocklin. Let's add those changes to this PR and then merge it so we can give intake a try!

martindurant:
@rabernat , so I should push to change to cat location to where the file will end up on github?

rabernat:
Since I see dask and bokeh being tweaked, is there any chance this will fix #236 (which is still live on the cluster)?

martindurant:
I'm afraid I don't know the status of that error.

rabernat:
Ok I'm going to merge this and then start building new docker images.

rabernat:
Oops, I missed this before merging:

Bokeh 0.12.16 is out. If I update to that, will it cause any conflicts with dask? (Our current version is 0.12.15dev1.)

martindurant:
https://travis-ci.org/dask/distributed/jobs/379395126#L651  seems to have passed tests with that bokeh. Not quite the same as running in production.

PR #251, opened 2018-05-11
Setup hpc

jhamman:
ping @kmpaul, @davidedelvento, and @mrocklin for review of hpc.rst
ping @rabernat for review of setup_guides/index.rst


rsignell-usgs:
The `hpc.rst` instructions worked perfectly for me on our Yeti Slurm cluster.   The only difference was that I launched the interactive session with: 
```
 salloc -p normal -A woodshole -n 1 -t 03:00:00
```
instead of 
```
qsub -I -l select=1:ncpus=4 -l walltime=03:00:00 -q regular
```

jhamman:
@rsignell-usgs - I've broken the setup instructions into two separate tracks, 1 using dask-jobqueue and the other using dask-mpi. I'm curious which you used?

docs/setup_guides/hpc.rst
```diff
@@ -38,27 +41,27 @@ Create a new conda environment for our pangeo work:
 ::
 
     conda create -n pangeo -c conda-forge \
-        python=3.6 dask distributed xarray jupyterlab mpi4py
+        python=3.6 dask distributed xarray jupyterlab mpi4py dask_jobqueue
```
jhamman: This will be possible soon: https://github.com/conda-forge/staged-recipes/pull/5844

rsignell-usgs:
@jhamman , I followed `dask-jobqueue`.     Also, I found it simpler/helpful after launching the interactive session to type this on the command line:
```
echo "ssh -N -L 8877:`hostname`:8877 -L 8878:`hostname`:8787 rsignell@yeti.cr.usgs.gov"
```
and then
```
jupyter lab --no-browser --ip=`hostname` --port=8877
```
Oh, and I did `conda env create -f pangeo_env.yml` with:
```
$ more pangeo_env.yml
name: pangeo
channels:
  - conda-forge
  - defaults
dependencies:
  - python=3.6
  - dask
  - distributed
  - xarray
  - jupyterlab
  - mpi4py
  - pip:
    - dask-jobqueue
```
Here's a notebook to demonstrate jobqueue working on USGS "Yeti" HPC:
https://gist.github.com/rsignell-usgs/21241994bb7df312f7d602814cd00529

Thanks for the great instructions!

docs/setup_guides/hpc.rst
davidedelvento: Here are my notes, by no means comprehensive:

* At the moment the start-notebook script on Cheyenne use the token, not the password. The person responsible for that script is working to change that, but maybe you want to mention the option to use the token to your users, especially because of what is mentioned below.

> a hashed password that encodes the same information, but is safe to include in a publicly accessible config file
* This is incorrect. The hash will only slow down potential attackers, not stop them. The config file should not be publicly accessible. 

* I noticed that your job scripts use **ompthreads**. I think that is needed only if your code is explicitly using OpenMP, otherwise you may be better served by just spawning the threads normally. But then we also enter into the discussion about thread/task pinning (aka processor binding) and that's a long one....

jhamman: @davidedelvento - thanks for giving this a read through. 

> maybe you want to mention the option to use the token to your users, especially because of what is mentioned below.

Until we have something functional that works with Cheyenne/Conda, I'm keen to keep this generic and to add a link to the jupyter documentation on this subject. I've kept this as HPC agnostic as possible with the hope that sys admins from groups like CISL would provide documentation specific to each machine. 

> The hash will only slow down potential attackers, not stop them. The config file should not be publicly accessible.

I will update accordingly. 

> I noticed that your job scripts use ompthreads...

I don't think this matters but it would be great if some CISL folks could evaluate this programatically. I can remove the `ompthreads` piece for now.

davidedelvento: You're very welcome @jhamman !

[token]
What to mention or not is your choice, obviously. But note that this is not Cheyenne-specific: tokens are the default setting for jupyter notebook now and that's why CISL supported them before passwords.

[ompthreads]
Well, that depends on what you intend to do. Perhaps @kmpaul knows both that and the way that OpenMP or regular threads work with PBS Pro?

shoyer:
Drive by comment about setting up a password for Jupyter: I'm a little surprised that this is the recommended way to get access to a remote notebook. In my experience, I always used ssh tunneling for this, and I assumed that a password is superfluous.

docs/setup_guides/hpc.rst
```diff
@@ -109,10 +111,114 @@ Copy that string into your ``jupyter_notebook_config.py`` config file
...
+is readable only by you. For more information on and other methods for
+securing Jupyter, check out
+`Securing a notebook server <http://jupyter-notebook.readthedocs.io/en/stable/public_server.html#securing-a-notebook-server>`__
+in the Jupyter documentation.
+
+------------
+
+From here, we have two options. Option 1 will start a Jupyter Notebook Server
+and manage dask and using the `dask-jobqueue`_ package. Option 2 will start a dask
+cluster using `dask-mpi` and will run a Jupyter server as part of the dask cluster.
```
mrocklin: Do we have a recommendation on which path is better to take?

docs/setup_guides/hpc.rst
```diff
@@ -109,10 +111,114 @@ Copy that string into your ``jupyter_notebook_config.py`` config file
...
+
+In our case, the Cheyenne super computer uses the PBS job scheduler, so typing:
+
+::
+
+    (pangeo) $ qsub -I -l select=1:ncpus=4 -l walltime=03:00:00 -q regular
+
+This will get us an interactive job on the `regular` queue for three hours. You
+may not see the `pangeo` environment anymore in your prompt, in this case, you
+will want to reactivate it.
```
mrocklin: Are interactive jobs like this standard across HPC systems?

docs/setup_guides/hpc.rst
```diff
@@ -109,10 +111,114 @@ Copy that string into your ``jupyter_notebook_config.py`` config file
...
+::
+
+    (pangeo) $ jupyter lab --no-browser --ip=`hostname` --port=8877
+    [I 13:36:52.298 LabApp] JupyterLab beta preview extension loaded from /home/username/miniconda3/envs/pangeo/lib/python3.6/site-packages/jupyterlab
+    [I 13:36:52.298 LabApp] JupyterLab application directory is /home/username/miniconda3/envs/pangeo/share/jupyter/lab
+    [I 13:36:52.321 LabApp] Serving notebooks from local directory: /home/username
+    [I 13:36:52.321 LabApp] 0 active kernels
+    [I 13:36:52.321 LabApp] The Jupyter Notebook is running at:
+    [I 13:36:52.321 LabApp] http://r8i4n0:8877/
+    [I 13:36:52.321 LabApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
```
mrocklin: You might want to avoid the logs here.  They might be a bit scary.

docs/setup_guides/hpc.rst
```diff
@@ -109,10 +111,114 @@ Copy that string into your ``jupyter_notebook_config.py`` config file
...
+
+*Note that we're also passing the 8878 port through so we can access the dask
+dashboard later.*
+
+
+Launch Dask with dask-jobqueue
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Most HPC systems use a job-scheduling system to manage job submissions and
+executions among many users. The `dask-jobqueue`_ is designed to help dask
```
mrocklin: the dask-jobqueue *package* is designed ...

docs/setup_guides/hpc.rst
```diff
@@ -312,23 +405,30 @@ Main script
     # >>> client = Client(scheduler_file='~/scheduler.json')
 
     rm -f scheduler.json
-    mpirun --np 6 dask-mpi --nthreads 6 --memory-limit 22e9 --interface ib0 --local-directory /glade/scratch/$USER
+    mpirun --np 6 dask-mpi --nthreads 6 \
+        --memory-limit 22e9 \
+        --interface ib0 \
+        --local-directory /glade/scratch/$USER
```
mrocklin: The local directory here will be an issue if someone copies this script naively.

docs/setup_guides/hpc.rst
```diff
@@ -348,3 +448,9 @@ And the second one a few times
 You can run this more times during your session to increase your
 allocation dynamically. You can also kill these jobs independently to
 contract your allocation dynamically and save compute time.
+
+.. _conda: https://conda.io/docs/
+.. _Jupyter: https://jupyter.org/
+.. _Dask: https://dask.pydata.org/
+.. _Cheyenne: https://www2.cisl.ucar.edu/resources/computational-systems/cheyenne
+.. _dask-jobqueue: http://dask-jobqueue.readthedocs.io
```
mrocklin: It would be good to point people to [Dask's HPC setup docs](http://dask.pydata.org/en/latest/setup/hpc.html) at some point.

docs/setup_guides/hpc.rst
```diff
@@ -1,26 +1,29 @@
...
+5. Connect to `Jupyter`_ and the `Dask`_ dashboard from your personal computer
 
-This document assumes that you already have an account on Cheyenne, and
-are comfortable using the command line. There is a
-`screencast <https://www.youtube.com/watch?v=7i5m78DSr34&feature=youtu.be>`__
-of this page.
+Although the examples on this page were developed using NCAR's `Cheyenne`_ super
+computer, the concepts here should be generally applicable to typical HPC systems.
+This document assumes that you already have an access to an HPC like Cheyenne,
+and are comfortable using the command line. It may necessaary to work with your
```
guillaumeeb: Typo in necessary here.

docs/setup_guides/hpc.rst
```diff
@@ -109,10 +111,114 @@ Copy that string into your ``jupyter_notebook_config.py`` config file
...
+Most HPC systems use a job-scheduling system to manage job submissions and
+executions among many users. The `dask-jobqueue`_ is designed to help dask
+interface with these job queing systems. Usage is quite simple and can be done
+from within your Jupyter Notebook:
+
+.. code:: python
+
+    from dask_jobqueue import PBSCluster
+
+    cluster = PBSCluster(processes=6, threads=4, memory="16GB")
```
guillaumeeb: Maybe put a more complete example here? Based on what you use on Cheyenne?

docs/setup_guides/hpc.rst
```diff
@@ -109,10 +111,114 @@ Copy that string into your ``jupyter_notebook_config.py`` config file
...
+*Note that we're also passing the 8878 port through so we can access the dask
+dashboard later.*
+
+
+Launch Dask with dask-jobqueue
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Most HPC systems use a job-scheduling system to manage job submissions and
+executions among many users. The `dask-jobqueue`_ is designed to help dask
+interface with these job queing systems. Usage is quite simple and can be done
```
guillaumeeb: Typo in "queing"

docs/setup_guides/hpc.rst
```diff
@@ -109,10 +111,114 @@ Copy that string into your ``jupyter_notebook_config.py`` config file
...
+For security reasons, we recommend making sure your ``jupyter_notebook_config.py``
+is readable only by you. For more information on and other methods for
+securing Jupyter, check out
+`Securing a notebook server <http://jupyter-notebook.readthedocs.io/en/stable/public_server.html#securing-a-notebook-server>`__
+in the Jupyter documentation.
+
+------------
+
+From here, we have two options. Option 1 will start a Jupyter Notebook Server
+and manage dask and using the `dask-jobqueue`_ package. Option 2 will start a dask
```
guillaumeeb: "manage dask and": and what?

docs/setup_guides/hpc.rst
```diff
@@ -169,7 +275,10 @@ Copy and paste the following text into a file, dask.sh:
     # >>> client = Client(scheduler_file='~/scheduler.json')
 
     rm -f scheduler.json
-    mpirun --np 12 dask-mpi --nthreads 6 --memory-limit 24e9 --interface ib0
+    mpirun --np 12 dask-mpi \
+        --nthreads 6 \
+        --memory-limit 24e9 \
+        --interface ib0
 
 This script asks for two nodes with 36 cores each. It breaks up each
```
guillaumeeb: In our PBS installation, such a script asks for two nodes with 72 cores each.

davidedelvento:
@shoyer Not having a password or a token might be ok if the server is your personal one which no one else can access. On a shared system such as an HPC environment, if you setup things like you said, **any** user of that system can effectively become you, read/write/delete your files regardless of permissions, burn out your whole allocation to run their job, abuse the system in many ways for which **you** will be held responsible by the site administrator, and probably even more damage.

docs/setup_guides/hpc.rst
davidedelvento: Depends what you mean by "like" and what you mean by "standard". Certainly almost all HPC systems in different places have some sort of interactive jobs somewhat similar to these. The details on how to start them, and where you can run them, and what environment they have is different from site to site.

docs/setup_guides/hpc.rst
davidedelvento: I agree. However sometimes they contain useful information for the user to see, especially if something goes wrong (and they are left in the dark, asking support for "it used to work, now it doesn't"). Perhaps we can grep-out only very standard lines and still print the rest, if any?

docs/setup_guides/hpc.rst
davidedelvento: Not sure if it's appropriate for other site, but `TMPDIR` at NCAR points to `/glade/scratch/$USER`. Alternatively, you can set somewhere (in the module?) an environmental variable to use here, something like `PANGEO_LOCALDIR`?

rsignell-usgs:
I'm just noticing that when I tunnel with: 
```
echo "ssh -N -L 8877:`hostname`:8877 -L 8878:`hostname`:8787 rsignell@yeti.cr.usgs.gov"
```
which generates:
```
ssh -N -L 8877:igskahcmgslih04.cr.usgs.gov:8877 -L 8878:igskahcmgslih04.cr.usgs.gov:8787 rsignell@yeti.cr.usgs.gov
```
the notebook on `localhost:8877` works fine, but the dashboard on `localhost:8787` gives "this site can't be reached"

rabernat:
That’s because the dashboard is on localhost:8878

docs/setup_guides/hpc.rst
jhamman: In my experience, almost all HPC systems have some way to get users interactive sessions on compute nodes. This helps get users off the log in nodes and provides users with an interactive environment to test jobs with the same environment that the batch queues get.

docs/setup_guides/hpc.rst
jhamman: I've shortened the log.

docs/setup_guides/hpc.rst
jhamman: Fixed above. Thanks for the catch.

docs/setup_guides/hpc.rst
jhamman: I added a section at the bottom for further reading.

rsignell-usgs:
@rabernat , jeez, yes, localhost:8878 works.  I was interpreting the mapping backwards.   🙃

jhamman:
Final comments here? If not, I plan to merge latter today. I hope others will continue to fine tune the HPC docs going forward.

davidedelvento:
Nothing to add from my side. You might want to ping @kmpaul one last time before merging.

kmpaul:
Sorry for the delay.  I'm on vacation right now.

I think this looks good.  There are a few "inconsistencies" in the documentation, but they can be fixed later (and I can do them next week when I'm back in the office).  The general content looks really good, though!  Thanks, for all the hard work!

PR #262, opened 2018-05-18
update worker template in notebook image

rabernat:
I forgot about this little step to point the notebook image to the worker image. (AFAIK, it is not documented anywhere!)


rabernat:
closing and updating #262 instead

PR #267, opened 2018-05-21
remove docker images from repo

rabernat:
They have moved to https://github.com/pangeo-data/helm-chart.


jhamman:
One thought I'm having: should we move the example notebooks to an independent repo?

mrocklin:
I'm in favor of this.  I also think that we should just copy this repo into
the user space on every load.

rabernat:
> I also think that we should just copy this repo into
> the user space on every load.

Is "this" referring to hypothetical new repository with example notebooks? If so, I agree.

mrocklin:
yes

PR #281, opened 2018-06-01
conda package is dask-jobqueue, not dask_jobqueue

rsignell-usgs:


rabernat:
Thanks for catching this!

Issue #232, opened 2018-05-01
List of current Pangeo deployments


jhamman:
We'd like to develop a current listing of Pangeo deployments. If you have deployed pangeo in one form or another, please speak up!

Information we'd like to know:

1. Cloud or HPC (or other)? Which system (e.g. Google Cloud, NCAR's Cheyenne)?
1. How are you deploying dask distributed (KubeCluster, dask-jobqueue, dask-mpi, etc.)?
1. How are you deploying Jupyter (Jupyterhub, single user)
1. Primary use (Tutorial, individual research, etc.)?

(I'll provide a few examples below, sharing a url to your jupyterhub deployment is not required).

xref #229 

---------

**pangeo.pydata.org** (Pangeo EarthCube)
1. Platform: Google cloud
1. Dask: KubeCluster
1. Jupyter: JupyterHub
1. Use: Exploratory deployment for Pangeo EarthCube project. Used for demos, tutorials, research.

**pangeo-aws.cloudmaven.org** (University of Washington)
*Note: this deployment will be taken down soon*
1. Platform: Amazon cloud
1. Dask:KubeCluster
1. Jupyter: JupyterHub
1. Use: Demo deployment / proof of concept for proposal.

**Cheyenne, Caldera, and Geyser** (NCAR)
1. Platform:  HPC 
1. Dask: dask-jobqueue and dask-mpi
1. Jupyter: Single user notebook servers (jupyterhub coming soon: #26 )
1. Use: Individual research

rabernat:
### [NASA Pleiades Cluster](https://www.nas.nasa.gov/hecc/resources/pleiades.html)

1. Platform: HPC
2. Dask: dask-mpi and [custom slurm launch scripts](https://github.com/pangeo-data/pangeo/pull/105)
3. Jupyter: single user notebook servers
4. Use: Ongoing ocean and climate data analysis

### [Columbia Habanero Cluster](https://confluence.columbia.edu/confluence/display/rcs/Habanero+HPC+Cluster+User+Documentation?src=contextnavpagetreemode)

1. Platform: HPC
2. Dask: dask-mpi and [custom slurm launch scripts](https://github.com/pangeo-data/pangeo/tree/master/utilities/habanero)
3. Jupyter: single user notebook servers
4. Use: Ongoing ocean and climate data analysis

tjcrone:
#### Lamont Real-Time Earth Pangeo Cluster
1. Platform: Azure cloud
[...]

PR #284, opened 2018-06-01
Add two deployments

spencerkclark:
@rabernat I hear that you will be giving a talk at GFDL this month.  In the spirit of #232, I figured it might be helpful to note that some users there are already using Pangeo software.  I've really been enjoying using dask-jobqueue on GFDL's Post-processing and Analysis cluster.  Many thanks to everyone who has worked on that.

For what it's worth I've also helped folks use dask-jobqueue on a Princeton research computing cluster, so I've added that too.

cc: @nathanieltarshish @jbusecke.

### DIFF
--- a/docs/data/deployments.yml
+++ b/docs/data/deployments.yml
@@ -98,3 +98,18 @@
   dask: dask-kubernetes
   jupyter: Jupyterhub
   kind: cloud
+
+- name: GFDL Post-processing and Analysis Cluster
+  description: HPC cluster running at the NOAA Geophyiscal Fluid Dynamics Laboratory
+  platform: HPC
+  dask: dask-jobqueue with a custom PBSCluster subclass
+  jupyter: Single user notebook servers
+  kind: hpc
+
+- name: Princeton Tiger Cluster
+  description: Research computing cluster running at Princeton University
+  link: https://researchcomputing.princeton.edu/systems-and-services/available-systems/tiger
+  platform: HPC
+  dask: dask-jobqueue
+  jupyter: Single user notebook servers
+  kind: hpc
--- a/docs/_templates/deployments.html
+++ b/docs/_templates/deployments.html
@@ -60,7 +60,7 @@
         <!-- <br> -->
         <li><em>Dask Deployment: {{ deployment.dask }}</em></li>
         <!-- <br> -->
-        <li><em>Jupyter Deploymentn: {{ deployment.jupyter }}</em></li>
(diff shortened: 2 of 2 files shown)


docs/data/deployments.yml
```diff
@@ -98,3 +98,18 @@
   dask: dask-kubernetes
   jupyter: Jupyterhub
   kind: cloud
+
+- name: GFDL Post-processing and Analysis Cluster
+  description: HPC cluster running at the NOAA Geophyiscal Fluid Dynamics Laboratory
+  platform: HPC
+  dask: dask-jobqueue with a custom PBSCluster subclass
```
jhamman: This interests me. If you have a moment, would you mind sharing a bit more about this. Either here, or preferably, on the dask-jobqueue issue tracker.

spencerkclark: Sure thing  - GFDL uses the [Moab Workload Manager](http://docs.adaptivecomputing.com/mwm/7-1-3/help.htm#cover_page.html) for batch job scheduling, which behaves very similarly to a standard PBS system (i.e. scripts take PBS directives etc.).  There are some differences, however, in the command names for submitting and cancelling jobs (see below).  As far as I can tell the only way to override those is by creating a subclass of PBSCluster.  In addition, I've found the need to override the function that parses the job ID, so that one can stop workers when desired (the default implementation does not parse it correctly).  The subclass is only a few lines:
```python
from dask_jobqueue import PBSCluster


class PPANCluster(PBSCluster):
    """A custom PBSCluster for use on the GFDL PPAN"""
    submit_command = 'msub'
    cancel_command = 'canceljob'

    def _job_id_from_submit_output(self, out):
        return out.strip()
```
I'm happy to report this in dask-jobqueue if you'd like.

guillaumeeb: That's interesting!

That would be nice to report this in dask_jobqueue. Maybe others are using Moab.

Could you share this as a pull request? Maybe renaming the class as MoabCluster.

spencerkclark: I'd be open to that.  Before I do that, I'll just note that I think Moab is proprietary.  Does that pose any problems?

jhamman: No problems that I can see.

spencerkclark: Sounds good.  Here is a link to the PR: dask/dask-jobqueue#66.

rabernat:
I know I'm late to the party, but thanks so much for sharing this!

PR #290, opened 2018-06-06
More use cases

jhamman:
cc @sridge


rabernat:
For the future, we should develop guidelines which clarify these formatting issues to make it easier to submit notebooks.

PR #292, opened 2018-06-06
new ssh notebook

rabernat:
This basically just an execution the notebook that's in the notebook docker image (with @willirath's fix for weighting).


docs/use_cases/index.rst
```diff
@@ -9,7 +9,7 @@ Physical Oceanography
 .. toctree::
    :maxdepth: 1
 
-   ocean_cm2.6_eddy_statistics
+   sea-surface-height
```
jhamman: do you want to remove the `ocean_cm2.6_eddy_statistics` notebook since  in this PR too?

rabernat: Yes. That notebook is out of date. I will be adding an updated one in a future PR.

PR #307, opened 2018-06-12
fix some minor typos

ocefpaf:
English is not my first language, so some of these correction may be wrong :grimacing:

### DIFF
--- a/docs/about.rst
+++ b/docs/about.rst
@@ -3,12 +3,12 @@
 About Pangeo
 ============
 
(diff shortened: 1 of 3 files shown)


rabernat:
Gah! So many typos! So embarrassing!

It is possible that some of this was written in a text editor without a spell checker. Also, we hustled very fast at the end to get it out there.

I see no reason not to merge these fixes right away. @ocefpaf your help is so appreciated!

PR #310, opened 2018-06-13
Remove pinning of JupyterHub image

mrocklin:
Merging soon if there are no comments.


jhamman:
Fine by me!

jacobtomlinson:
I seem to remember pinning this when the default was still v5 to pull it up to the newer version. I don't see why this can't be removed now.

PR #318, opened 2018-06-18
Added Rob Hetland to people.yml

hetland:


rabernat:
Welcome Rob! We would love to learn about how you are using (or interested in using) pangeo.

PR #325, opened 2018-06-27
Pangeo-Binder

jhamman:
I've been working on getting a Pangeo-like Binder deployment up and running that will support using dask-kubernetes.

Some key points:

- docker builds the notebook image, can include dask-kubernetes
- working on exposing the Google Container Registry docker image path to notebook image environment/dask-kubernetes (xref: https://github.com/jupyterhub/kubespawner/pull/193)
- not yet fully working prototype is running here: http://104.198.48.82

I'm hoping this allows us to have click-and-run examples on our website. It also has the nice advantage of providing a mechanism to share the notebook/worker images which may be useful in the traditional jupyterhub setting.

I think @martindurant, @mrocklin, @tjcrone, and @choldgraf will be interested in following the progress here.


mrocklin:
Do we have the ability to control which environments are allowed or not?  Or will this be fully public access?  In my mind we would ideally be able to separately control who can provide new environments and who can use the cluster.

mrocklin:
Also, thoughts on how to handle worker environments?  Do we just use the same docker image?

jhamman:
moving development to stand alone repo: https://github.com/pangeo-data/pangeo-binder

PR #326, opened 2018-06-29
Fix typo in HPC setup

lesteve:
While I was at it I noticed a wrong port in the jupyter + dask-jobqueue doc.

### DIFF
--- a/docs/setup_guides/hpc.rst
+++ b/docs/setup_guides/hpc.rst
@@ -122,7 +122,7 @@ in the Jupyter documentation.
 From here, we have two options. Option 1 will start a Jupyter Notebook server
 and manage dask using the `dask-jobqueue`_ package. Option 2 will start a dask
 cluster using `dask-mpi` and will run a Jupyter server as part of the dask cluster.
-We generally recommend starting with Option 1, espcially if you will be working
+We generally recommend starting with Option 1, especially if you will be working
 interactively, unless you have a reason for managing the job submission scripts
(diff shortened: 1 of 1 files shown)


docs/setup_guides/hpc.rst
```diff
@@ -181,7 +181,7 @@ Now, connect to the server using an ssh tunnel from your local machine
 
 ::
 
-    $ ssh -N -L 8877:r8i4n0:8877 -L 8878:r8i4n0:8787 username@cheyenne.ucar.edu
+    $ ssh -N -L 8877:r8i4n0:8877 -L 8787:r8i4n0:8787 username@cheyenne.ucar.edu
```
lesteve: Not 100% sure about this one, but 8787 instead of 8787 feels like a typo, since everywhere else 8878 is used.

Maybe a good idea would be to use ports that are more different for jupyter notebook and the dashboard port.

lesteve:
On top of the jupyter and the dashboard port being hard to distinguish, there is no consistency between the jupyter + dask-jobqueue (notebook 8877, dashboard 8878) and the jupyter + dask-mpi (notebook 8888, dashboard 8787) doc.

I can change the ports to be more dissimilar and harmonize it between dask-jobqueue/dask-mpi if you think it is worth it.

lesteve:
Looking a bit closer, it looks like the doc have some assumption, basically that the dashboard is on the default port for the dask-jobqueue doc and that the notebook server is on the default port for dask-mpi, so maybe harmonization is not possible between the two.

Issue #309, opened 2018-06-12
xesmf included in pangeo.pydata.org ?


naomi-henderson:
Hi all,
I have put together a use-case for reproducing the canonical ICCP-AR5 Figure12.5 using the CMIP5 multi-model surface temperature data.  I have converted the needed netcdf files to zarr and uploaded to  pangeo.pydata.org. I just noticed that the xesmf packages has not been included, so I can't do the fast and easy spatial regridding provided by xesmf to combine all models on the same grid. See:
[use case on pangeo.pydata.org](http://pangeo.pydata.org/user/naomi-henderson/lab/tree/CMIP5-MakeFigure12.5.ipynb).   
Any chance we can add xesmf to the standard packages?

Here is the figure this notebook can reproduce:

![wgi_ar5_fig12-5](https://user-images.githubusercontent.com/31460695/41320286-2d2626de-6e6d-11e8-9504-c79999b4a8a7.jpg)

rabernat:
This is a great point @naomi-henderson. (FYI, we can't follow the link <http://pangeo.pydata.org/user/naomi-henderson/lab/tree/CMIP5-MakeFigure12.5.ipynb>. If you would like to share a notebook from pangeo, your best bet right now is to download it and upload it as a gist.)

We have discussed xesmf use cases in #197, but your example is much more concrete. It would be a great example to highlight the capabilities of pangeo to the climate community.

The main roadblock right now is pangeo-data/helm-chart#29, which is holding us up from adding new packages to pangeo.pydata.org. Hopefully that will be resolved very soon, and we can move ahead with this.

naomi-henderson:
Ah, thanks @rabernat -  I will wait for this to be sorted out.  I will also need xESMF to finish the atmospheric moisture budget use case. The ability to do regridding is essential in dealing with the CMIP5 (and soon to come CMIP6) data.

mrocklin:
xesmf is now on the stock software environment on Pangeo.  I believe that this is now resolved.  I'm curious to see it used.

cc @JiaweiZhuang

naomi-henderson:
@mrocklin fantastic!  So far so good, xesmf is happily calculating the weight files and reusing them when appropriate. Thanks much

rabernat:
I wonder how xesmf will work with distributed. Will it scatter the weight matrix to the workers?

JiaweiZhuang:
[...]

PR #341, opened 2018-07-26
add gmet zarr dataset to intake cat

jhamman:
xref: https://github.com/pangeo-data/pangeo-example-notebooks/pull/5

### DIFF
--- a/gce/catalog.yaml
+++ b/gce/catalog.yaml
@@ -3,7 +3,7 @@ plugins:
     - module: intake_xarray
 sources:
   newmann_netcdf:
-    description: newmann meteo ensemble in netcdf
+    description: Newman met ensemble in netcdf
     driver: netcdf
     args:
       urlpath: '/gcs/newmann-met-ensemble-netcdf/conus_ens_00*.nc'
@@ -11,7 +11,7 @@ sources:
       concat_dim: ensemble
       chunks: {'time': 50}
   newmann_zarr:
-    description: newmann meteo ensemble in netcdf
+    description: Newman met ensemble in zarr format (9 ens members)
     driver: zarr
     args:
       storage_options:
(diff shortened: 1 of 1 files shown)


martindurant:
I am +1 for this.
However, the full dataset contains rather a lot of files and is slow to load the metadata. The chunksize of 14MB is OK order-of-magnitude but perhaps would be better at ~100MB. Perhaps this data-set also needs the metadata treatment in https://github.com/pangeo-data/pangeo/issues/309 , and a general agreement on how to cope with the many-files problem for the future (half-million files in this case).

jhamman:
@martindurant - I can update the chunksize in the dataset if that would be useful.

martindurant:
I certainly think that would help, and ~> 100MB is a good place to aim, perhaps bigger as the total dataset increases.

jhamman:
Right. I may remove the .zarr suffix though. Thoughts?

mrocklin:
> Perhaps this data-set also needs the metadata treatment in #309

How does one do this?  Do we need a best practices doc?  Should this be
upstreamed to XArray in some way?

martindurant:
@mrocklin , right, it's strictly ad-hoc, we didn't come to  a conclusion on how to consolidate the metadata in a way to causes least friction. My simplistic solution was to put a file `.zmetadata` in the top-level with the contents of all of the rest of the `.z*` files in the hierarchy; then `zarr.Group` looks for this key in the mapping. Following the discussion on metadata over on zarr, it could, for example, have been an additional entry in `.zattrs` in the top-level, and so stay within the spec.

martindurant:
(but yes, it could have been a PR to xarray, that's part of the conversation in #309)

martindurant:
> I may remove the .zarr suffix though on the bucket. Thoughts?

I mildly side with keeping it, for people that browse the file hierarchy in the bucket.

mrocklin:
OK.  I'd like to make sure that if this is important that it doesn't get
lost.  It looks like nothing significant has happened on that issue for
about a month.  @martindurant do you have thoughts on the right way to come
to some long-term conclusion on this issue?  Should it be elevate to some
upstream issue tracker?

martindurant:
We have https://github.com/zarr-developers/zarr/pull/268 , which is also getting old.

alimanfoo:
Happy to pick that discussion up again on the Zarr side if that would help.

jhamman:
I rechunked the dataset. We're down to 16k chunks and up to ~80mb/chunk.

Issue #366, opened 2018-08-20
Feedback on Deploying Pangeo on the Cloud doc


guillaumeeb:
Finally I gave it a quick try this afternoon, here are my remarks/issues FWIW.

Disclaimer: I used Google Cloud Shell to perform the commands, as in France (I don't know in other countries), companies love to give windows only personal computer, and its difficult to find a linux server open to the internet.

From step one to step three, it was very straightforward. Some minor comments:
- It would be nice to add more details on what we are doing, for example in step two, explaining that we are creating a cluster with a default pool of two nodes (for the hub?), and a worker pool that can scale from 0 to 100 nodes (for single user notebook and dask workers ?).
- You must have appropriate rights on the project to issue some commands, especially step three. If you're owner of the google project, obviously it's OK.
- `gcloud config set container/new_scopes_behavior true` is probably not about being authenticated.

At step four, I felt a little left over: what must I modify in this `jupyter_config.yaml` file?
- Is `EXTRA_PIP_PACKAGE` proposed value OK?
- Should I use `GCSFUSE`, and point to pangeo-data bucket or mine?
- `customPodHook` looks kind of a hack, and do I really need to add a logo file in it?
- And most of all: what should I put in `loadBalancerIP`? Should I inspect Kubernetes cluster with the command provided in step 5? Should I use one of the two nodes from default pool external IP (at that point I just realized that default nodes does have external IP, not like a default compute instance)? Should I start a Load balancer on the Cloud? Ingress?

Okay we'll see later, I try some values (leave a lot of default except IP), and then go to step five. But no luck here, the command did not work:
The trace I get (can this be related to cloud shell use?):
```
 "pangeo" has been added to your repositories
Hang tight while we grab the latest from your chart repositories......
Skip local chart repository...
Successfully got an update from the "pangeo" chart repository...
Successfully got an update from the "stable" chart repository
Update Complete. ⎈ Happy Helming!⎈
Error: failed to download "pangeo/pangeo" (hint: running `helm repo update` may help)
```

So more generally, I would suggest the following:
- Add more details on what we are doing and why we are doing it, even if it's not the primary goal of this doc,
- Tell what must be changed/adapted in `jupyter_config.yaml`, special attention to give to the IP
- Keep the bash script somewhere in github, so that users can just git clone them and edit them easily.  https://github.com/pangeo-data/pangeo/blob/master/gce/setup.sh is quite old doesn't reflect what is on setup quide.

[...]

Issue #373, opened 2018-08-27
refactoring pangeo.pydata.org


rabernat:
As we discussed at the developer meeting, the time may have come to refactor pangeo.pydata.org.

Let's discuss how we would like our future services to look. I think we are all in agreement that, for non-persistent demos, we definitely want to move to @jhamman's binder.pangeo.io solution.

But what about the "production" services. There are some people (including myself, @chiaral, @chiara7, and who knows who else) who are using pangeo.pydata.org for actual science.

I personally really like @mrocklin's idea of refactoring pangeo.pydata.org into several more specific services. For example:
- ocean.pangeo.io: all of my high resolution ocean models and physical oceanography stuff
- cmip.pangeo.io: focused on CMIP5 / CMIP6 / large ensemble climate models
- forecast.pangeo.io: atmospheric reanalysis and forecast
- hydro.pangeo.io: @jhamman's and @rsignell-usgs's hydrologic models and datasets
- solar.pangeo.io? astro.pangeo.io?

Each cluster could have a distinct:
- environment with domain-specific packages
- set of examples
- data catalog (although all data would be available to all clusters)
- dedicated administrator

The advantage of this approach is that it might catalyze collaboration around specific scientific problems. But in order for that to happen, we would still need a way to make it easier to share examples with the group.

mrocklin:
> But in order for that to happen, we would still need a way to make it easier to share examples with the group.

When we interact with a new commuity we'll be giving them a cookie-cutter repository template.  That repository will have an `examples` directory that they will be encouraged to fill with notebooks.  Given how well this has worked in the original deployment I expect that other communities will continue this tradition.  As long as this is in a uniform place (which is likely given that we're providing a cookie cutter) then presumably we can scape these directories and place them into either documentation, or provide a binder-able version of the same repository.

rabernat:
Good points Matt. But I worry that the complexity of adding new notebooks to the examples repo will exclude some users. Currently that process is something like:
- Copy an existing example to a new location (this is how most people will probably start their work)
- Modify it into something new that you want to share
- Download the notebook by right-clicking in the file browser
- Create a PR to the appropriate example-notebooks repo
- Wait for PR to be merged
- Users have to start their jupyter pod again to see the examples refresh

[...]

PR #378, opened 2018-09-05
update cloud setup guide

guillaumeeb:
Some updates of the setup guide as discussed in #366.

I'm unsure in what Kubernetes pool the single user notebook are spawned, and may have written something false.


guillaumeeb:
Of course all proposed changes here are just a quick fix before a better solution based on terraform or something like that is proposed, as discussed in https://github.com/pangeo-data/pangeo/issues/373#issuecomment-416725442.

guillaumeeb:
I think that the update here is somewhat stable and the cluster creation procedure is complete. We could maybe get this in after some reviews while we wait for something more automatized as is being discussed in #384.

rabernat:
Ok LGTM

PR #386, opened 2018-09-14
new SOSE use case

rabernat:
An example from the Southern Ocean State Estimate


mrocklin:
I can't say I understand much of what's happening here.  But it sure does look nice!

rabernat:
> I can't say I understand much of what's happening here. But it sure does look nice!

That's a good thing! It means we are doing actual science. 😉

rabernat:
The one problem is that the holoviews dynamic map does not make it into the rendered output. There is probably a way to make it at least render one image.

mrocklin:
You can't use github, which blocks active javascript.  You need to use nbviewer (or anaconda.org)

http://nbviewer.jupyter.org/github/pangeo-data/pangeo/blob/45706493a8136e285538c7e8d86a5fd7cd7228a4/use_cases/physical-oceanography/SOSE.ipynb

rabernat:
He the notebook is getting rendered to html by [sphinx_nbexamples](https://pypi.org/project/sphinx-nbexamples/). And the holoviews cell doesn't make it.

I wonder if @Chilipp might know now how to resolve this.

mrocklin:
Also cc @jcrist who might want to look at some of the applications where dask.array is being used in order to get a sense of potental places where atop fusion might help.

rsignell-usgs:
There is something in the notebook like “save widget state”. Does that help?

-Rich

jbednar:
In case it helps clarify any of the above:
- A DynamicMap does include one frame of the visualization when exported, so if nothing is showing up, it's usually because JS is being blocked (as for github or some doc building systems).
- You can export more frames if you convert the DynamicMap to a HoloMap by sampling multiple specific frames; once it's been made concrete like that it can be exported so that the sliders will work.
- Saving the widgets in Jupyter won't help; I believe that's a hack specifically for ipywidgets.
- You can set holoviews to use PNG as its output format, which will make the images show up everywhere.  But if you do that, then they won't be interactive anywhere, and they will also be *really* slow to create, because to get PNG we have to fire up a separate web browser to do the rendering.  

What would be nice is if we could have some easy way to build a PNG version of an existing page, without changing it but somehow generating an .ipynb where the JS has been rendered into PNG.  The components for doing so all already exist, but I don't know how hard it would be to make it happen.  I do think it would be useful, which would allow notebooks to be compiled to PDF via LaTeX. @philippjfr, any ideas?

philippjfr:
Is there somewhere I could look at a build of the website that was rendered using sphinx_nbexamples? I have a slight suspicion that this could be related to an old version of ``pyviz_comms``, so it might be worth trying to upgrade that to 0.6.0 if it isn't already.

>What would be nice is if we could have some easy way to build a PNG version of an existing page, without changing it but somehow generating an .ipynb where the JS has been rendered into PNG. The components for doing so all already exist, but I don't know how hard it would be to make it happen.

HoloViews does allow doing this already by declaring the ``display_formats`` in the extension, e.g.:

```python
hv.extension('bokeh', display_formats=['png', 'html'])
```

However I suspect that sphinx_nbexamples will still try to use the JS.

jlstevens:
As you are using bokeh in this notebook, I think it would be best to find out what is stopping the JS from executing. Is there anything reported in the web console?

rabernat:
> Is there somewhere I could look at a build of the website that was rendered using sphinx_nbexamples?

The current version of the Pangeo uses it: http://pangeo.io/use_cases/index.html
But none of those examples have holoviews. Once we merge this PR, this notebook will go online and you can see it there.

It sounds like this holoviews issue is potentially resolvable but not immediately. So I am going to merge this and just accept that the holoviews cell will not show up for now.

philippjfr:
Okay, once it's built I'll probably be able to provide concrete guidance on how to fix it. Will be happy to submit a PR.

philippjfr:
Could someone with travis credentials try adding this environment variable to the travis config and then triggering a rebuild?

HV_DOC_HTML='true'

rabernat:
Environment variable added to Travis.  No change in output as far as I can tell: http://pangeo.io/use_cases/physical-oceanography/SOSE.html#visualize-some-data

philippjfr:
It looks like something is stripping out all script tags. I'll play around with sphinx_nbexamples a bit and report back.

philippjfr:
There's a section about including bokeh plots in the [sphinx_nbexamples docs](https://sphinx-nbexamples.readthedocs.io/en/latest/getting_started.html#including-bokeh). That may be a good start.

Chilipp:
Hey!

> It looks like something is stripping out all script tags. I'll play around with sphinx_nbexamples a bit and report back.

This is definitely true and the reason is the conversion from `.ipynb` to `.rst` using `nbconvert` (sphinx-nbexamples converts all notebooks to rst using `nbconvert` and includes the rst files in the docs, in principle: `jupyter nbconvert SOSE.ipynb --to=rst`). 
The same happens if you convert it to `.md`, but it works for HTML. I wonder whether is this comes from what is reported in https://github.com/ioam/holoviews/pull/2574. As far as I got, the hack mentioned in this PR is only implemented for the HTML conversion, but not for rst or markdown.

rsignell-usgs:
I've been using Google Calendar to organize my calendar, find interesting  
events, and share my schedule with friends and family members. I thought  
you might like to use Google Calendar, too.

rsignell@usgs.gov recommends that you use Google Calendar.

To accept this invitation and register for an account, please visit:  
[https://www.google.com/calendar/render?cid=reply%2B001c92d8062ac1612080e9e666eb6a993c0f54c67dfb16eb92cf0000000117b38c9692a169ce157a3482@reply.github.com]

Google Calendar helps you keep track of everything going on in
your life and those of the important people around you, and also help
you discover interesting things to do with your time.

rsignell-usgs:
(I just deleted the strange google calendar invite that appeared to come from my usgs.gov email.  I don't see any sent messages these, so I'm still trying to figure out how that happened)

PR #389, opened 2018-09-18
pin jhub version

rabernat:
this is necessary with the latest pangeo helm chart


rabernat:
I'm gonna try deploying with #390 but without this and see if it works. If so, I will close this.

rabernat:
Ok, I just deployed #390 and it works.

Issue #391, opened 2018-09-19
Weekly checkin meeting?


mrocklin:
Do we want to have a short weekly checkin meeting to synchronize work and issues happening around Pangeo development?  This might help to align efforts and also form a stronger social base.

guillaumeeb:
That definitly sounds good to me! 

For example I'm currently experiencing cloud deployment when I find free time, but I'm not convinced I use and update the correct procedure, see #378. It seems there is some activity on this with helm chart or jupyterhub guide. And there's also an experiment with terraform and maybe other tools mentioned by @jacobtomlinson.

I'm also trying to work on a basic satellite image processing example with Dask and Xarray, there may be some existing material around here...

rabernat:
Thanks for suggesting this Matt. It's a good idea. My schedule is very tight this semester, but I will try to make it as much as possible.

My preferred time would be Wednesday morning.

FYI @jhamman is currently on vacation.

jhamman:
+1 on this idea. Wednesday mornings would work for me (or probably any other morning). Just keep in mind that I'm on Pacific Time so preferably nothing earlier than 8 PT.

jacobtomlinson:
8PT would be 15GMT so that's fine with me. I am about to disappear on paternity leave at some point in the next few weeks, so probably best to get at least one in before I go. 

I would suggest we aim to have one tomorrow?

guillaumeeb:
Wednesdays are not ideal for me, I have to leave work at 17:20, sot 15:20 UTC due to too many paternity leaves already taken 🙂. I would prefer tuesday or thursday where 15UTC is generally fine for me.

Not available tomorrow, sorry for that, but it's certainly fine to do that without me, just keep everyone posted of the discussions!

kmpaul:
Wednesday mornings usually work well for me, so I'll put my vote in for that.

mrocklin:
I suggest that we meet regularly at 8am Wednesday mornings.  

I apologize that this doesn't work well for you @guillaumeeb .  It was unlikely that we would find a time that worked well for everyone.

mrocklin:
Sorry, 8am US Pacific, 9am US Mountain, 11am US Eastern, 15:00 UTC

jacobtomlinson:
Sounds good to me. Are we checking in at https://appear.in/pangeo

guillaumeeb:
[...]

PR #397, opened 2018-09-26
added embedded meeting notes

rabernat:
This PR adds a new page to the website with a simple embed of the weekly meeting notes (see #391). Might be nice to keep this process as open as possible.

What do people think? Appropriate to put out in public or no?

### DIFF
--- a/docs/meeting-notes.rst
+++ b/docs/meeting-notes.rst
@@ -1,0 +1,8 @@
+.. _meeting-notes:
+
+Pangeo Weekly Meeting Notes
+===========================
+
+.. raw:: html
+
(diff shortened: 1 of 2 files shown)


mrocklin:
Fine by me to keep them public.  I would just link to a view-only version
of the google doc to avoid having to copy-paste every week.

rabernat:
> I would just link to a view-only version of the google doc to avoid having to copy-paste every week.

That's exactly what this is.

mrocklin:
Ah!  My mistake.  Should have looked more closely :)

guillaumeeb:
> What do people think? Appropriate to put out in public or no?

Wasn't there, and it was very usefull for me to read this. Not sure if this will be useful to many people, but I believe it's great to have all this in public.

docs/meeting-notes.rst
```diff
@@ -0,0 +1,8 @@
+.. _meeting-notes:
+
+Pangeo Weekly Meeting Notes
+===========================
```
jhamman: Since this meeting is open to anyone, it may be worth putting the schedule and call in details here. Then we can close the GH issue (#391).

PR #406, opened 2018-10-01
Add GitHub organizations to whitelist

dsludwig:
cc @rabernat @mrocklin


gce/secret-config.yaml
```diff
@@ -8,6 +8,13 @@ jupyterhub:
       clientId: "2cb5e09d5733ff2e6ae3"
       clientSecret: SECRET
       callbackUrl: "http://pangeo.pydata.org/hub/oauth_callback"
+      org_whitelist:
+        - pangeo-data
+        - pydata
+        - dask
+        - xarray-dev
+        - jupyterhub
+        - informatics-lab
```
mrocklin: I'm happy with all members of all of these orgs having access.  Hopefully this is only a short-term fix anyway until we refactor things.

PR #407, opened 2018-10-01
update github org members when people.yml is updated

jhamman:
This is a rough, but working, sketch of how we can keep http://pangeo.io/collaborators.html#people in sync with the `@pangeo-data/pangeo-members` team and the pangeo-data github organization. Both of which can still be managed manually.

### DIFF
--- a/docs/tools/member.py
+++ b/docs/tools/member.py
@@ -40,8 +40,6 @@ def main(access_token, people_yml):
 
 if __name__ == "__main__":
 
-    print(list(os.environ.keys()), flush=True)
-
     parser = argparse.ArgumentParser(description='Add users to GitHub Org/Team')
     parser.add_argument('--access_token', required=True, type=str,
                         help='GitHub Access Token')
--- a/docs/tools/member.py
+++ b/docs/tools/member.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import os
 import argparse
 import yaml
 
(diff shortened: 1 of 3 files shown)


rabernat:
This is really clever.

However, it does seem sketchy. Do we really want travis interacting directly with github? Could someone manipulate this to take over the pangeo organization?

Based on my read of the code, I don't see much that could go wrong. You are using `safe_load` on the yaml file, so there isn't much chance of injecting malicious code or something. But I am not an expert on these things.

docs/tools/member.py
```diff
@@ -0,0 +1,50 @@
...
+    # pangeo-members = 2180253
+    team = org.get_team(2180253)
+
+    # get existing GitHub members
+    existing_org_logins = list(member.login for member in org.get_members())
+    existing_team_logins = list(member.login for member in team.get_members())
+
+    # add those members not yet in the GitHub org
+    for member in github_ids:
+        user = g.get_user(member)
```
rabernat: do we need to catch an error here if `member` is not actually a valid github user?

jhamman:
Another approach would be to build a Github App that does this. The main problem with this approach is that I don't know where to start.

mrocklin:
How serious of a problem is this?  We could also handle things the old
fashioned way.  In the model of the refactor other groups will handle
github teams on their own.  I'm inclined to ignore this problem.

jhamman:
@mrocklin - I see this as enough of a problem that we should be exploring options. Some reasons for pursuing a solution in this area.

- We want to be able to share a list of collaborators - this is useful both in terms of building community and encouraging buy-in.
- We want to be able to communication with these people - unless they are in a github org or listserve, we don't have anyway to batch ping these people. Github provides nice tools for managing notification levels.
- We want to be able to use the pangeo org as a blanket whitelist for things like pangeo.pydata.org - though we're working on a refactor, this fact doesn't go entirely away. 

Some reasons why the status quo is not going to work:

- manually keeping these lists in sync is unlikely to succeed
- we don't have another way to add people to the github org without an admin

PR #408, opened 2018-10-01
Simplify setting up a password.

Carreau:
Note that this will store the password in the _json_ config file, likely
`~/.jupyter/jupyter_notebook_config.json` that _takes precedence_ over
the py file (for set options).

The first time you open a jupyter notebook _without passing the token in
the URL_ it also let you set a password from the UI.

-- 
This make setting a password dramatically simpler, but I wasn't sure if old manual way of setting pwd should be removed.

### DIFF
--- a/docs/setup_guides/hpc.rst
+++ b/docs/setup_guides/hpc.rst
@@ -69,6 +69,14 @@ Configure Jupyter
 (If you don't plan to use Jupyter notebooks then you can safely skip
 this section.)
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


Carreau:
> I hope to see more of you around here in the future :)

Likely, my new job seem to imply being close to kubernetes, clusters, and users.

So far looking at how we can get kernels started in different pods than the notebook server, and I was looking into whether pangeo was doing it.

jhamman:
@Carreau - we're not doing the different kernels in different pods but that is an interesting idea.

PR #425, opened 2018-10-10
big catalog update

rabernat:
This PR adds many ocean datasets to our intake catalog which have been there for a while but have not been easily discoverable.

### DIFF
--- a/gce/catalog.yaml
+++ b/gce/catalog.yaml
@@ -28,10 +28,122 @@ sources:
         token: anon
         access: read_only
       urlpath: 'gcs://pangeo-data/gmet_v1.zarr'
-  sea_surface:
+  sea_surface_height:
     description: sea-surface altimetry data from The Copernicus Marine Environment
+    metadata:
+      url: 'http://marine.copernicus.eu/services-portfolio/access-to-products/?option=com_csw&view=details&product_id=SEALEVEL_GLO_PHY_L4_REP_OBSERVATIONS_008_047'
+      tags:
+        - ocean
+        - satellite
     driver: zarr
     args:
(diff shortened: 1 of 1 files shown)


rabernat:
The travis build failed with this error:
```
Exception occurred:
  File "/home/travis/miniconda/envs/docs-environment/lib/python3.6/site-packages/intake/catalog/local.py", line 155, in __init__
    self._plugin = global_registry[driver]
KeyError: 'netcdf'
The full traceback has been saved in /tmp/sphinx-err-27qz74o5.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 1
```

However, this doesn't seem related to my PR, since I didn't add any datasets with the netcdf driver. That was already in there:
https://github.com/pangeo-data/pangeo/blob/b60e932f7307c4d484337d4da1e3842037362d51/gce/catalog.yaml#L5-L12

martindurant:
Passing now. Needed a re-release of Intake. Not sure what was going on, should check that the netcdf sources actually load, rather than getting skipped.

rabernat:
cool, thanks martin!

Anyone else have comments before I merge?

(Ideally we should do this for all the datasets, but I just want to get the ball rolling.)

martindurant:
As far as I can tell, it look fine. The proof is in testing within a running instance.

Do you make use of the metadata tags? Perhaps those can be rendered on the catalog HTML page.

rabernat:
Yes, eventually I would like to use the metadata tags for various purposes.

PR #454, opened 2018-10-30
provide IDs and links to individual datasets

rabernat:
I would like to be able to link to specific entries in the catalog. I believe this does that.

### DIFF
--- a/docs/catalog.rst
+++ b/docs/catalog.rst
@@ -14,8 +14,8 @@ These datasets are directly accessible from the Google Cloud Pangeo deployment:
 
       <tr>
         <td>
+          <div id="{{ name }}">
           <ul class="list-inline">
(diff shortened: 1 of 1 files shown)


rabernat:
I don't know. I haven't actually tried building it locally.

martindurant:
OK, I only just twigged what you were doing here. This will be more useful as the catalog gains more entries. 
Might there be a reason at some point to split the cats into separate files and a hierarchy?

rabernat:
Yes. I eventually imagine many sub-catalogs curated by the different sub-groups within pangeo, which are then aggregated into one big catalog.

rabernat:
I guess I'll just merge an see how it looks.

martindurant:
Anchors work, but there is excessive underlining.

PR #537, opened 2019-01-19
Add gpm_imerg_early, trmm_3b42rt, cgiar_pet, hydrosheds

davidbrochart:


davidbrochart:
There is a problem with the hydrosheds dataset, as it looks like the Rasterio driver doesn't work with a remote path: https://github.com/ContinuumIO/intake-xarray/issues/31

jhamman:
Thanks @davidbrochart - I think its okay to merge this in its current state. We'll follow along with the intake-xarray issue.

PR #540, opened 2019-01-23
Fix hydrosheds dataset

davidbrochart:
Fixed using https://github.com/ContinuumIO/intake-xarray/issues/31.


jhamman:
thanks @davidbrochart!

PR #402, opened 2018-09-27
Small fixes in the cloud setup guide for GCE

arokem:
This fixes a few small typos and formatting issues in the current format of the webpage. I also added links directly into this GH repo for the files that are mentioned throughout. It might be best to remove the verbatim copying of these files into this page so that we don't have to keep checking that they are consistent? 

One small issue that I ran into is that when I tried to run `4_upgrade_helm.sh`, I got the error message: 

    Error: failed to download "pangeo/pangeo" (hint: running `helm repo update` may help)

No such error with `5_upgrade_helm.sh`. I think that this was already captured elsewhere, but couldn't find that.

PR #547, opened 2019-02-07
Fix rst problems in cloud.rst

lesteve:
**Edit**: added following link + snapshot for clarification:
Looking at the `pangeo.io` doc (https://pangeo.io/setup_guides/cloud.html):

and seeing things like:
pangeo

It looks like something went slightly wrong in the rst at one point. You can see that in https://github.com/pangeo-data/pangeo/blob/master/docs/setup_guides/cloud.rst (search for `` `_ `` to look for problems).

I looked for similar problems with `git grep` and there were none left.

### DIFF
--- a/docs/setup_guides/cloud.rst
+++ b/docs/setup_guides/cloud.rst
@@ -107,9 +107,8 @@ create a cluster with the components that Pangeo requires:
     notebook servers.
   - An auto scaling node pool for Dask workers.
 
-This script is availabe in pangeo/gce/setup-guide as
-` ``1_create_cluster.sh`` <https://github.com/pangeo-data/pangeo/blob/master/gce/setup-guide/1_create_cluster.sh>`_,
-so you can use it directly.
+This script is availabe in pangeo/gce/setup-guide as `1_create_cluster.sh`_, so
+you can use it directly.
 
 .. Note::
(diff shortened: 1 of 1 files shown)


arokem:
Looks like this would eliminate the links to the scripts on github?

lesteve:
I was looking at the `pangeo.io` doc (https://pangeo.io/setup_guides/cloud.html):

and seeing things like:
pangeo

But looking at the readthedocs doc (https://pangeo.readthedocs.io/en/latest/setup_guides/cloud.html), the links seems fine.

Not sure the differences in how the two docs are generated ...

(from what I gathered I also learned a magical reST, or maybe sphinx, way of linking to a file plus including its content)

arokem:
I think that the RTD version is out-dated. I don't think there's any magic there, it's just that the scripts are written out verbatim into that version of the docs. The goal of #402 was to avoid the duplication. Obviously, it seems to have introduced other issues... But unless I am missing something, I don't think that the current PR is a good solution to these issues, because it eliminates the links to GH.

arokem:
Oh no - I am wrong. The links are defined at the bottom. 

+1 for this change!

guillaumeeb:
I had no Idea there was still something on readthedocs. This does not sound normal.

Maybe open another issue and cc @rabernat.

PR #560, opened 2019-02-22
Add script to tunnel clean chrome-flavored browser

willirath:
This adds a script that has been in use at GEOMAR for quite some time.

It simplifies all the tunneling effort in, e.g., <http://pangeo.io/setup_guides/hpc.html> to 

```shell
start_tunneled_chrome.sh  <username>@<host-behind-the-firewall>  <url-i-got-from-jupyter>
```


willirath:
The current version should just work on Linux / MacOS with Chrome or Chromium installed.

There's a few things that would benefit from polishing:

1. Handling the SSH tunnel with a socket rather than the current 
https://github.com/pangeo-data/pangeo/blob/a4fcee1039f49f62eeba01cc78631c8208b51bb6/utilities/tunneling/start_tunneled_chrome.sh#L24
and hope for the browser to create some traffic before the tunnel dies after 60 seconds.

2. Add the default paths for Windows and run all this from a git-bash.exe which has a proper SSH client.

willirath:
3. Use `mktemp -d` for creating and tearing down a temporary user dir?

willirath:
To test this (@rabernat and @jhamman ?), you could 

curl -L https://raw.githubusercontent.com/pangeo-data/pangeo/a4fcee1039f49f62eeba01cc78631c8208b51bb6/utilities/tunneling/start_tunneled_chrome.sh | bash -s < http://www.displaymyhostname.com/

willirath:
4. Add google-domains to the bypass list.  Otherwise, on HPC with a strict firewall, chrome trying to phone home to see if there's an update available, or if there are settings you might wish to sync, would show up in the firewall logs a lot.

rabernat:
Willi - Sorry this has gone without a response for a long time.

Your script looks incredibly useful. The hesitation is that I'm not sure that the Pangeo repo is the right place to host this sort of material.

What if you just created a gist and then providing a link from the pangeo docs? That would accomplish the same purpose but relieve the pangeo repo of the burden of maintaining your script.

PR #592, opened 2019-04-11
Minor fixes on hpc setup guide for cheyenne

leosiqueira:
updating conda command usage for newer version where ``source activate`` was replaced with ``conda activate`` and added account name for interactive qsub.

### DIFF
--- a/docs/setup_guides/hpc.rst
+++ b/docs/setup_guides/hpc.rst
@@ -158,11 +158,24 @@ in the Jupyter documentation.
 
 Finally, we may want to configure dask's dashboard to forward through Jupyter.
 This can be done by editing the dask distributed config file, e.g.:
-``.config/dask/distributed.yaml``. In this file, set:
+``.config/dask/distributed.yaml``. By default, when ``dask.distributed`` or
+``dask-jobqueue`` is first imported, it places a file at ``~/.config/dask/distributed.yaml``
+with a commented out version. You can create this file and do this first import by simply 
+
+::
+
+    python -c 'from dask.distributed import Client'
+
+In this ``.config/dask/distributed.yaml`` file, set:
 
 .. code:: python
 
-    diagnostics-link: "/proxy/{port}/status"
+  #   ###################
+  #   # Bokeh dashboard #
+  #   ###################
+  #   dashboard:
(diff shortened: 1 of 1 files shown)


docs/setup_guides/hpc.rst
```diff
@@ -79,7 +97,8 @@ this section.)
 .. note::
 
    When using recent Jupyter iteration the following section can be replaced by simply invoking the command::
```
guillaumeeb: So we should make commands wit plural?

Reading the whole documentation, I think we don't want to ignore the section completly, as the two last steps should be done either case. Maybe we should add a subsection for setting the password?

docs/setup_guides/hpc.rst
```diff
@@ -170,15 +189,15 @@ In our case, the Cheyenne super computer uses the PBS job scheduler, so typing:
 
 ::
 
-    (pangeo) $ qsub -I -l select=1:ncpus=4 -l walltime=03:00:00 -q regular
+    (pangeo) $ qsub -I -A account -l select=1:ncpus=4 -l walltime=03:00:00 -q regular
```
guillaumeeb: -A account is required for Cheyenne, but not necessarily with other systems. This is not mandatory in CNES for example.

docs/setup_guides/hpc.rst
leosiqueira: Agreed, and the two last steps are indeed important in any case.

docs/setup_guides/hpc.rst
leosiqueira: It is indeed not required in many other systems, but at the beginning the page says the examples were developed using NCAR's Cheyenne and the sentence just before this says "In our case, the Cheyenne ... uses the PBS job scheduler, so typing:".

guillaumeeb: OK, fine by me.

rabernat:
Is this ready to merge?

leosiqueira:
I think so for now. Guillaume is suggesting that we also should consider changing the config section layout at some point.

guillaumeeb:
So let's merge this, we'll improve the rest later on.

PR #596, opened 2019-04-17
Add stub for french 2019 meeting

willirath:
See https://github.com/pangeo-data/pangeo/issues/493


rabernat:
Thanks for this @willirath! I'll add this for now.

Going forward I think we want one top-level page for all our meetings.

guillaumeeb:
Thank you both!

PR #616, opened 2019-05-08
[DOC] Install guide for Azure cloud platform

DPeterK:
Adds a guide for installing Pangeo on an Azure kubernetes cluster.

### DIFF
--- a/docs/setup_guides/azure.rst
+++ b/docs/setup_guides/azure.rst
@@ -237,7 +237,8 @@ the IP address of the Pangeo proxy:
   kubectl get service proxy-public --namespace=pangeo
 
 Note that this service can take a long time to start up, so you may need to wait
-a while for the IP address of the Pangeo proxy to be displayed.
+a while for the IP address of the Pangeo proxy to be displayed. The output of the
+above command will read ``<pending>`` while the service is starting up.
 
 Once the service has started up, navigate to the ``EXTERNAL-IP`` address listed
 in the output of the above command in your web browser.
--- a/docs/setup_guides/azure.rst
+++ b/docs/setup_guides/azure.rst
@@ -79,7 +79,7 @@ https://docs.microsoft.com/en-us/azure/virtual-machines/windows/sizes.
 .. note::
   RBAC (Role-Based Access Control) is enabled by default on Kubernetes services set up using
   the Azure CLI. If you need to change this default behaviour, you can specify the
-  ``disable-rbac`` flag when creating your Kubernetes cluster.
+  ``--disable-rbac`` flag when creating your Kubernetes cluster.
 
 The ``az aks create`` command above assumes that you have already set up a resource group
 to deploy your Kubernetes service into. If you have not already set up a resource group,
@@ -219,7 +219,7 @@ Test install
 To test that Pangeo has installed successfully on your Kubernetes service, find
 the IP address of the Pangeo proxy:
 
-.. code-block::
+.. code-block:: bash
 
   kubectl get svc proxy-public --namespace=pangeo
 
--- a/docs/setup_guides/azure.rst
+++ b/docs/setup_guides/azure.rst
@@ -1,0 +1,249 @@
+Installing Pangeo on Azure
+--------------------------
+
+This guide takes you through the steps necessary to install Pangeo on Microsoft's Azure Cloud Platform.
+We'll make use of Azure's Kubernetes as a Service offering, called AKS (Azure Kubernetes Service),
+for installing Pangeo on Azure.
+Documentation on AKS can be found here: https://docs.microsoft.com/en-gb/azure/aks/.
+
+.. Note::
+  This guide lays out only the fundamental steps required to install Pangeo on Azure AKS.
+  Further work, for example to secure your cluster, is highly advised but not directly
+  covered here.
+
+
+Step One: Build Kubernetes service
+==================================
+
+The first step to installing Pangeo on Azure is to set up a Kubernetes service
+that can be used to run Pangeo. This can be done either by using the web interface
+or by using the Azure commandline interface (CLI). These are both practical options,
+so we'll cover each one in turn.
+
+Using the web interface
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To use the Azure web interface you must first have a Microsoft account that you
+can use to log into the Azure web interface. If you have an existing Microsoft
+account (for example, a hotmail.com or outlook.com email address) then you can use
+that, or you can create a new account.
+
+Once you have logged into the Azure web interface, navigate to Kubernetes services
+and click the blue Add logo in the top left. This will display the Create Kubernetes cluster
+wizard.
+
+Work through the wizard customising the Kubernetes service to be created as you
+see necessary (though the defaults are largely reasonable). In the last step before
+the cluster is created a validation process is run, ensuring that the customisations you have
+made will produce a working cluster. At this step you can also download a
+template file to make reproducing or automating cluster creation simpler in the future.
+
+Autoscaling
+```````````
+
+One benefit of the web interface is that we can easily create an AKS resource that implements
+autoscaling via virtual nodes (see https://docs.microsoft.com/en-us/azure/aks/virtual-nodes-portal
+for further details on this concept). Virtual nodes are still a preview feature in Azure, so some
+limitations currently apply. To enable virtual nodes you need to have followed the setup steps in the
+link above and also be using an Azure region where nodes are supported.
+
+With both of these requirements met, you can enable virtual nodes for autoscaling your Kubernetes
+service. In the Scale tab, ensure that both the Virtual Nodes and VM scale sets selectors are
+set to 'Enabled'.
+
+
+Using the Azure CLI
+~~~~~~~~~~~~~~~~~~~
+
+Instructions for downloading and installing the Azure CLI on major Operating Systems
+can be found at: https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest.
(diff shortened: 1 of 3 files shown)


rabernat:
Thanks so much for this contribution @DPeterK!

Could someone else who has worked with Azure give this a quick review? @tjcrone?

DPeterK:
Thanks @rabernat! I think @jacobtomlinson was also keen to take a look.

docs/setup_guides/azure.rst
```diff
@@ -0,0 +1,249 @@
...
+To use the Azure web interface you must first have a Microsoft account that you
+can use to log into the Azure web interface. If you have an existing Microsoft
+account (for example, a hotmail.com or outlook.com email address) then you can use
+that, or you can create a new account.
+
+Once you have logged into the Azure web interface, navigate to Kubernetes services
+and click the blue Add logo in the top left. This will display the Create Kubernetes cluster
+wizard.
+
+Work through the wizard customising the Kubernetes service to be created as you
```
jacobtomlinson: For beginners we may want to be more explicit here. Either the defaults are completely acceptable or we should suggest some specific changes.

docs/setup_guides/azure.rst
```diff
@@ -0,0 +1,249 @@
...
+see necessary (though the defaults are largely reasonable). In the last step before
+the cluster is created a validation process is run, ensuring that the customisations you have
```
jacobtomlinson: `customisations` should we standardize on the American Dictionary?

docs/setup_guides/azure.rst
```diff
@@ -0,0 +1,249 @@
...
+for installing Pangeo on Azure.
+Documentation on AKS can be found here: https://docs.microsoft.com/en-gb/azure/aks/.
+
+.. Note::
+  This guide lays out only the fundamental steps required to install Pangeo on Azure AKS.
+  Further work, for example to secure your cluster, is highly advised but not directly
+  covered here.
+
+
+Step One: Build Kubernetes service
```
jacobtomlinson: A screenshot might be nice?

docs/setup_guides/azure.rst
```diff
@@ -0,0 +1,249 @@
...
+To create a basic Kubernetes service using the Azure CLI:
+
+.. code-block:: bash
+
+  az aks create \
+    --resource-group $RESOURCE_GROUP_NAME \
+    --name $AKS_RESOURCE_NAME \
+    --kubernetes-version 1.12.6 \
+    --node-count 1 \
+    --node-vm-size Standard_B8ms
```
jacobtomlinson: Should we explain what `Standard_B8ms` is? and why we've chosen it?

docs/setup_guides/azure.rst
```diff
@@ -0,0 +1,249 @@
...
+we've also asked for a medium-sized VM to host the node rather than the default.
+You can specify any VM name listed in the links from this page as the value to this key:
+https://docs.microsoft.com/en-us/azure/virtual-machines/windows/sizes.
+
+.. note::
+  RBAC (Role-Based Access Control) is enabled by default on Kubernetes services set up using
+  the Azure CLI. If you need to change this default behaviour, you can specify the
+  ``disable-rbac`` flag when creating your Kubernetes cluster.
+
+The ``az aks create`` command above assumes that you have already set up a resource group
```
jacobtomlinson: Should we throw in the command for making a resource group?

jacobtomlinson: Ah I see you've done this already!

docs/setup_guides/azure.rst
```diff
@@ -0,0 +1,249 @@
...
+and click the blue Add logo in the top left. This will display the Create Kubernetes cluster
+wizard.
+
+Work through the wizard customising the Kubernetes service to be created as you
+see necessary (though the defaults are largely reasonable). In the last step before
+the cluster is created a validation process is run, ensuring that the customisations you have
+made will produce a working cluster. At this step you can also download a
+template file to make reproducing or automating cluster creation simpler in the future.
+
+Autoscaling
```
jacobtomlinson: Should we mention that traditional autoscaling is not possible with the GUI and needs to be done by the CLI?

docs/setup_guides/azure.rst
```diff
@@ -0,0 +1,249 @@
...
+Now we can move onto installing Pangeo on our Kubernetes service. This can be done
+as follows:
+
+.. code-block:: bash
+
+  helm repo add pangeo https://pangeo-data.github.io/helm-chart/
+  helm repo update
+  helm upgrade --install --namespace pangeo pangeo pangeo/pangeo -f pangeo.yaml
+
+The contents of the ``pangeo.yaml`` file came from the Zero to Jupyterhub guide.
```
jacobtomlinson: I feel like we are skipping over the ` pangeo.yaml` file a bit here. We should probably provide an example.

docs/setup_guides/azure.rst
```diff
@@ -0,0 +1,249 @@
...
+~~~~~~~~~~~~
+
+To test that Pangeo has installed successfully on your Kubernetes service, find
+the IP address of the Pangeo proxy:
+
+.. code-block::
+
+  kubectl get svc proxy-public --namespace=pangeo
+
+Navigate to the ``EXTERNAL-IP`` address listed there in your web browser.
```
jacobtomlinson: This can take a while to be populated. Might be worth mentioning.

betatim:
What is your plan for maintaining these docs?

The reason I ask is that the Zero-to-JupyterHub guide already contains a section on setting up a Kubernetes cluster and the related tooling on Azure. I think it would make sense to pool our resources for creating documentation instead of having very similar docs in two places (and requires twice as much effort to keep updated).

The pangeo on GKE guide already links out to the z2jh guide so maybe that can be a model to follow?

DPeterK:
@betatim thanks for pointing this out - there is a lot of duplicated content between this and the existing z2jh guide, although the z2jh guide is more in-depth. My intention here was to provide a setup document that exactly tracked what I did that was similar in content to the existing Pangeo on GCE setup guide.

I'm quite happy to remove the content in the section on setting up an AKS cluster from this guide and just point to the existing z2jh guide, unless others feel that having explicit content here on setting up the cluster would be more beneficial?

jacobtomlinson:
This is an interesting discussion. Pangeo is a derivative of Z2JH and there are many situations where we end up referring back to the Z2JH docs. 

However in order to build a Pangeo stack we run through very similar steps to Z2JH but just with different config.

docs/setup_guides/azure.rst
DPeterK: I feel like I meant to add more content here, but didn't...

betatim:
I don't know what the best solution here is.

Having one document that lists all steps is super useful. One complaint we have received about the Zero-to-BinderHub guide is that it links out to other guides for some steps which results in you jumping around a lot. Which is hard if you are new to the whole thing.

One approach I've started liking (but not a lot of feedback on it yet) is to have a "checklist" style guide that is duplicated. And then one central extensive guide that explains why you are running the commands in the checklist and goes more in depth.

So yeah, I think the answer is "it is complicated".

jacobtomlinson:
> jumping around a lot
This is my concern.

> guide that is duplicated
This is also a concern.

> it is complicated
Yup 🤣

rabernat:
Thanks everyone for this useful discussion. I think we have identified the key tension here: we don't want to just duplicate the z2jh guides, but we also want to provide a more-or-less complete guide for incoming new users.

One solution could be to use templates. We could imagine taking a page like this ([step-zero-azure-autoscale.rst]https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/source/microsoft/step-zero-azure-autoscale.rst) and refactoring it into a jinja2 template, with each of the steps as its own block, plus variables for certain items (machine type, etc.) The main "vanilla" z2jh guide could use the template in its unmodified, default form. Pangeo setup guides (or other similar derivatives) could overwrite / modify elements as they see fit. This way we would get the best of both worlds: we could all publish complete, standalone guides while minimizing duplication.

guillaumeeb:
FWIW, I really like the Pangeo setup guide on GCE, because it gives all that is needed to deploy a Pangeo platform. It refers to z2jh as a reference, and people can go there to go deeper. z2jh doc is really usefull, but less easy to follow as it covers a lot of possible configurations.

I see the Pangeo setup guide a bit like @betatim describes the checklist approach. OK it's a little more than that, but I think that's the idea.

I did'nt read the Azure guide in detail, but I see that there is a `Step 3: Install Pangeo` item. This part should be identic to the [helm use on GKE](http://pangeo.io/setup_guides/cloud.html#step-five-deploy-helm-chart), doesn't it? Maybe we could use templating here?

guillaumeeb:
And I forgot to be polite, please excuse me, thanks for the really important and detailed work here @DPeterK !!

guillaumeeb:
Yeah, I understand what you mean @jacobtomlinson! Let's build on the really good work from @DPeterK, this could and will help several people, no need to over do it now.

By the way Jacob, where is the install guide for AWS 😁 ?

betatim:
Please do merge. I just wanted to plant the seed of this idea not solve it now as it is a tricky problem.

jacobtomlinson:
@guillaumeeb you make a good point! Pete is definitely showing me up here 😂. However he also built our new AWS cluster recently so maybe he can write that too ;).

jacobtomlinson:
Failures due to #621. Merging anyway.

DPeterK:
> he also built our new AWS cluster recently so maybe he can write that too

Beware the sloping shoulders! (Also, thanks for the merge 😁)

PR #638, opened 2019-05-20
Correct links for data catalog

fmaussion:
I've looked for other references and this seems to be the only one.


rabernat:
Fantastic, thank you!

PR #694, opened 2019-08-20
Update data.rst to remove GCFaces

cspencerjones:
gcsfs.GCSMap has been retired. I suggest replacing with 'fsspec.get_mapper'


jhamman:
Thanks @cspencerjones!

PR #704, opened 2019-08-28
Dev website

jsignell:
This ended up different than how I originally imagined, but hopefully it is simpler. 

Some constraints that I hadn't considered:
 1. You can't deploy from a fork (this makes sense if you think about it from a security perspective)
 2. Deploying to another repo is annoying

### What this PR does:
 1. When you push to the `dev` branch, the docs are built as usual, and then the `dev` directory gets added to the gh-pages branch (you should be able to see it there now.
 2. That dir isn't avialable in the nav for the pangeo website but you can access it at this url: https://pangeo.io/dev/

### Some downsides:
 1. I suspect that the dev dir on gh-pages will get over-written whenever someone merges to master. But actually that might be a good thing since we don't want master to get ahead of dev.
 2. The dev branch might get stale.

### How this would be used in practice:
Once a PR has been deemed reasonable, we could merge it to dev as a form of staging, inspect it at https://pangeo.io/dev/ and then if it looks reasonable, merge dev to master and delete the dev branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,5 +22,5 @@ after_script:
   - if [[ "${TRAVIS_BRANCH}" == "master" ]]; then
       doctr deploy . --built-docs docs/_build/html;
     else
-      doctr deploy . --built-docs docs/_build/html --no-require-master --deploy-repo pangeo-data/pangeo-dev-website;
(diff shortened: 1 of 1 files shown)


jsignell:
I just rebased this off master to avoid conflicts

PR #728, opened 2019-10-12
Update sea-surface-height.ipynb

AyushJoker:
count changed


jhamman:
Hi @AyushJoker - Thanks for opening a pull request. I don't think we need to merge this change though. The execution count of a ipynb document is not important in the context of our documentation build. Thoughts?

PR #769, opened 2020-03-19
Replace pangeo.pydata.org with hub.pangeo.io

jrbourbeau:
This PR changes references to `pangeo.pydata.org` (which has been closed) to `hub.pangeo.io`

### DIFF
--- a/docs/architecture.rst
+++ b/docs/architecture.rst
@@ -90,7 +90,7 @@ Cloud
 Commercial Cloud Computing offers the scientific community a fundamentally new
 model for doing data science. A specific focus of Pangeo is the development of
 scalable, flexible, and configurable cloud deployments. Our flagship deployment
-is `pangeo.pydata.org <http://pangeo.pydata.org>`__ running on Google Cloud
+is `hub.pangeo.io <https://hub.pangeo.io>`__ running on Google Cloud
 Platform.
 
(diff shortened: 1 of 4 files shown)


tjcrone:
hub.pangeo.io appears to point to the dev deployment. Is this the "flagship" deployment? For some reason I thought it was ocean, but I'm not sure about that. @rabernat can you please confirm that we want to point users to dev in these pages?

jrbourbeau:
> Is this the "flagship" deployment?

I thought it was based on the description of `hub.pangeo.io` in https://pangeo.io/deployments.html#deployments), but I could be wrong about that. Happy to make updates accordingly

tjcrone:
Okay cool thank you @jrbourbeau. I don't maintain the GCP deployments so we just need a quick weigh in from @jhamman or @rabernat on this. I think they probably want to point to either dev or ocean, but I'm not sure. Thanks for this helpful PR.

jhamman:
I think this is fair change. ocean.pangeo.io does get more traffic but is (obviously) intended for a specific audience. 

Ultimately, I want to shut down hub.pangeo.io and move ocean.pangeo.io to a new name (gcp-uscentral1.pangeo.io). This is going to require some coordination between @rabernat, myself, and others.

PR #789, opened 2020-09-29
Document setting environment variables on Dask cluster

TomAugspurger:


TomAugspurger:
@rsignell-usgs IIRC you had a use-case for this? Apologies I didn't get around to it till now (when I had a use-case for it :smile:)

If this is for the USGS deployment, the changes are at https://github.com/pangeo-data/pangeo-cloud-federation/pull/763/files. That also has some formatting changes. The main change is adding

```python
               return Options( 
                   ...
                   Mapping("environment", {}, label="Environment Variables"), 
```

and the line `                       "environment": options.environment, ` in the `option_handler`.

PR #797, opened 2020-10-18
Update hpc.rst

rsignell-usgs:
* move the jupyterlab extensions to the Jupyter section
* remove the old complicated way of specifying the password
* remove the manual install of the nbserverproxy jupyter serverextension because this is already installed by conda


rsignell-usgs:
replaced by #826

PR #800, opened 2020-11-03
(WIP) Start batch workflow docs

TomAugspurger:
This adds a section on batch workflows.

FireShot Capture 016 - Pangeo Cloud — Pangeo documentation -

I've taken a few liberties (the default image doesn't have MLFlow, papermill, or papermill-mlflow-handler, but we can add those to the docker image to make the docs correct).


TomAugspurger:
The approach here didn't pan out. https://kbatch.readthedocs.io/en/latest/ is an alternative option.

PR #811, opened 2020-12-10
Test forked example

charlesbluca:
The PR workflow works on branches within the base repo, but what about forks?


charlesbluca:
So on forked PRs, we just want the most basic tests running (just building the site).

PR #820, opened 2021-03-13
Fix link to software environment section

jrbourbeau:
This PR removes an extra trailing underscore which causes this link to not be rendered correctly (see screenshot below)

Screen Shot 2021-03-12 at 9 22 05 PM


jhamman:
Thanks @jrbourbeau!

PR #826, opened 2021-04-06
 hpc instructions to use mamba and small fixes

### LINKED ISSUE #825: Modify HPC instructions to use base env for jupyterlab
I would like to modify the [Pangeo software environment instructions for HPC](https://pangeo.io/setup_guides/hpc.html#installing-a-software-environment) to be more aligned with what I recommend people do, but wanted to check to see if others agree.  

The basic procedure is:
* install miniconda  (unchanged)
* specify conda-forge as the only channel in the config
* install mamba (for faster dependency solving)
* update all packages in the base env
* in the base env, install the jupyterlab stuff (jupyterlab, jupyter-server-proxy, dask-labextension, ipywidgets, nb_conda_kernels, nodejs)
* create one or more custom environments that contain other packages like (dask, dask-jobqueue, xarray, holoviz, ipykernel, ipywidgets and extra packages). 
* pick a random port to use on the server instead of assuming 8888 will be available. 
* launch jupyterlab from the base environment on that port

Would it be okay for me to submit a PR with these changes, or is this too opinionated (or someone has a better best practice)?

rsignell-usgs:
fixes #825

### DIFF
--- a/docs/setup_guides/hpc.rst
+++ b/docs/setup_guides/hpc.rst
@@ -181,37 +181,59 @@ From here, we can start jupyter. The Cheyenne computer administrators have
 developed a `start-notebook <https://www2.cisl.ucar.edu/resources/computational-systems/cheyenne/software/jupyter-and-ipython#notebook>`__
 utility that wraps the following steps into a single execution. You should
 check with your system administrators to see if they have something similar.
-If not, you'll need to take the following steps:
-
-Copy this line into your terminal. It will echo a command you'll want to use
-later.
-
-::
-
-    (pangeo) $ echo "ssh -N -L 8888:`hostname`:8888 $USER@cheyenne.ucar.edu"
-    ssh -N -L 8888:r8i4n0:8888 username@cheyenne.ucar.edu
-
-Now we can launch the notebook server:
+If not, you can easily create your own start_jupyter script.  In the script below, we choose a random port
+on the server (to reduce the chance of conflict with another user), but we use port 8889 on the client, as port 8888 is 
+the default client port if you are running Jupyter locally.  We can also change to a starting directory:
 
 ::
 
-    (pangeo) $ jupyter lab --no-browser --ip=`hostname` --port=8888
+    (pangeo) $ more ~/bin/start_jupyter 
+    cd /home/data/username
+    JPORT=$(shuf -i 8400-9400 -n 1)
+    echo ""
+    echo ""
+    echo "Step 1: Wait until this script says the Jupyter server"
+    echo "        has started. "
+    echo ""
+    echo "Step 2: Copy this ssh command into a terminal on your"
+    echo "        local computer:"
+    echo ""
+    echo "        ssh -N -L 8889:`hostname`:$JPORT $USER@poseidon.whoi.edu"
+    echo ""
+    echo "Step 3: Browse to https://localhost:8889 on your local computer"
+    echo ""
+    echo ""
+    sleep 2
+    jupyter lab --no-browser --ip=`hostname` --port=$JPORT
+
+Now we can launch the Jupyter server:
+
+::
+
+    (pangeo) $ ~/bin/start_jupyter
+    
+    To connect to this server, paste this line into a terminal on your client (your local machine)
+    after you see the line "Jupyter Server ... is running... ":
+    ssh -N -L 8889:pn009:8537 username@poseidon.whoi.edu
     ...
-    [I 13:36:52.321 LabApp] The Jupyter Notebook is running at:
-    [I 13:36:52.321 LabApp] http://r8i4n0:8888/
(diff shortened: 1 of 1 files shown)


docs/setup_guides/hpc.rst
```diff
@@ -34,12 +34,19 @@ After you have logged into your HPC system, download and install Miniconda:
...
+
+Before creating your environment, update your conda package manager with packages 
+from the conda-forge channel instead of the default channel and install Mamba, 
+which works like conda but is written in C++ and therefore creates environments faster. 
 
 ::
     
-    conda update conda
+    conda config --add channels conda-forge --force
+    conda config --remove channels defaults --force
```
guillaumeeb, line 46: I'm not sure why you need to use only conda-forge? This is to ensure the best compatibility between all the dependencies. This config might prevent using other channels like pytorch, don't this? 
But at this point if a user needs other channels he knows what he's doing, so maybe that's OK.

docs/setup_guides/hpc.rst
```diff
@@ -175,37 +181,47 @@ From here, we can start jupyter. The Cheyenne computer administrators have
...
+    JPORT=$(shuf -i 8400-9400 -n 1)
+    echo ""
+    echo ""
+    echo "Step 1: Wait until this script says the Jupyter server"
+    echo "        has started. "
+    echo ""
+    echo "Step 2: Copy this ssh command into a terminal on your"
+    echo "        local computer:"
+    echo ""
+    echo "        ssh -N -L 8889:`hostname`:$JPORT $USER@poseidon.whoi.edu"
```
guillaumeeb, line 201: `poseidon.whoi.edu`: maybe replace with a generic name.

guillaumeeb (commented):
Just two remarks, otherwise this looks really good!

docs/setup_guides/hpc.rst
```diff
@@ -175,37 +181,47 @@ From here, we can start jupyter. The Cheyenne computer administrators have
...
+    [I 2021-04-06 06:33:57.962 ServerApp] Jupyter Server 1.5.1 is running at:
+    [I 2021-04-06 06:33:57.962 ServerApp] http://pn009:8537/lab     
+    [I 2021-04-06 06:33:57.963 ServerApp]  or http://127.0.0.1:8537/lab
+    [I 2021-04-06 06:33:57.963 ServerApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
 
-You'll want to change the details in the command above but the basic idea is
-that we're passing the port 8888 from the compute node `r8i4n0` to our
-local system. Now open http://localhost:8888 on your local machine, you should
-find a jupyter server running!
+Just follow the Steps 1,2,3 printed out by the script to get connected.
```
andersy005, line 224: 📣 Shameless plug 📣 ;) : You could replace all of this with a **single command** via [jupyter-forward](https://github.com/NCAR/jupyter-forward) - Jupyter-forward launches Jupyter Lab over SSH (on any remote machine) with a single command !

Here are some examples:

- Launching Jupyter Lab on a remote host’s login node: https://jupyter-forward.readthedocs.io/en/latest/examples.html#launching-jupyter-lab-on-a-remote-host-s-login-node
- Launching Jupyter Lab on a remote host’s compute node: https://jupyter-forward.readthedocs.io/en/latest/examples.html#launching-jupyter-lab-on-a-remote-host-s-compute-node

rsignell-usgs: @andersy005 , what do you think about leaving these steps in here, and then saying that many of them can be automated by jupyter-forward as long as you are not on windows?   

I'm hesitant to just replacing this with jupyter-forward recommendation until https://github.com/NCAR/jupyter-forward/issues/81 is figured out.

docs/setup_guides/hpc.rst
ocefpaf: Any kind of channel mixing is problematic. We always recommend to use only 1 and, if it is extremely necessary, add the extra channel in the env file configuration. However, the ideal of keeping defaults and conda-forge is usually problematic, specially without strict channel policy. That is b/c those two channels overlap a lot. It is not like getting pytorch from another channel, it is like mixing from two different sources for all packages.

The other advantage, besides avoiding compatibility issues, is to speed up the solver time by reducing the amount of packages it has to work with.

rsignell-usgs: Thanks for the clarification Filipe!  This has been a moving target for many of us, so good to know that one channel is the way to go now!

docs/setup_guides/hpc.rst
rsignell-usgs: @andersy005 , what do you think about just merging this and following it with something about jupyter forward as another PR?

guillaumeeb:
I'll just go ahead and merge this one!

PR #840, opened 2021-06-07
Update Azure deployment guide

TomAugspurger:
This rewrites the Azure deployment guide. I'll detail the major changes below, but I wanted to ask up front: does pangeo want to be in the business of maintaining these guides, especially with 2i2c taking over maintenance roles? cc @tjcrone, @rabernat, and maybe @choldgraf. In terms of actual maintenance cost I'm happy to keep the Azure guide up to date.

Selfishly, I just want a place to link to for "How do I replicate the [Planetary Computer Hub](https://planetarycomputer.microsoft.com/docs/overview/environment/) in my Azure subscription?". That could live at https://planetarycomputer.microsoft.com/docs/concepts/computing if pangeo doesn't want to maintain these guides anymore.

Last bit of house-keeping: why not just use zero to jupyterhub? Two reasons:

1. There's a bit of awkwardness in "go to this other guide, but stop before deploying jupyterhub, since we're actually using this other helm chart". Not the worst in the world, but it's nice to read everything in one place (at the cost of maintenance).
2. https://zero-to-jupyterhub.readthedocs.io/en/latest/kubernetes/microsoft/step-zero-azure.html has some recommendations around networking and AKS auth I need to investigate. *Maybe* just an issue with AKS best-practices having moved on since that was written.

---

Onto the changes: I don't think the diff is too meaningful. But at a high level, the guide has two sections: A "simple" deployment, just to make sure that the tools work, and a "recommended" deployment. I've also included a sample `Makefile` and `config.yaml`, to satisfy people who just want to get a cluster up and running ASAP, and don't care about the background.

The details below has the rendered HTML as a png

<details>

pangeo-azure

</details>

### DIFF
--- a/docs/setup_guides/azure.rst
+++ b/docs/setup_guides/azure.rst
@@ -1,267 +1,416 @@
 Installing Pangeo on Azure
---------------------------
+==========================
 
-This guide takes you through the steps necessary to install Pangeo on Microsoft's Azure Cloud Platform.
-We'll make use of Azure's Kubernetes as a Service offering, called AKS (Azure Kubernetes Service),
-for installing Pangeo on Azure.
-Documentation on AKS can be found here: https://docs.microsoft.com/en-gb/azure/aks/.
+In this guide, you'll deploy a Pangeo-style JupyterHub on Microsoft Azure. This will deploy a multi-user JupyterHub enabled with Dask for scalable computing.
 
-.. Note::
-  This guide lays out only the fundamental steps required to install Pangeo on Azure AKS.
-  Further work, for example to secure your cluster, is highly advised but not directly
-  covered here.
+We'll use
 
+* `Azure Kubernetes Service <https://docs.microsoft.com/en-us/azure/aks/intro-kubernetes>`__ (AKS), Azure's managed Kubernetes service
+* The `daskhub Helm Chart <https://github.com/dask/helm-chart/tree/main/daskhub>`__, an easy way to install JupyterHub and Dask-Gateway
 
-Step One: Build Kubernetes service
-==================================
+We describe two deployments, a :ref:`simple` and a :ref:`recommended`. If you're new to Azure, Kubernetes, or JupyterHub, then you should try the simple deployment to verify that the basics work, before moving on to the more advanced recommend deployment.
 
-The first step to installing Pangeo on Azure is to set up a Kubernetes service
-that can be used to run Pangeo. This can be done either by using the web interface
-or by using the Azure commandline interface (CLI). These are both practical options,
-so we'll cover each one in turn.
-
-Using the web interface
-~~~~~~~~~~~~~~~~~~~~~~~
-
-To use the Azure web interface you must first have a Microsoft account that you
-can use to log into the Azure web interface. If you have an existing Microsoft
-account (for example, a hotmail.com or outlook.com email address) then you can use
-that, or you can create a new account.
-
-Once you have logged into the Azure web interface, navigate to Kubernetes services
-and click the blue Add logo in the top left. This will display the Create Kubernetes cluster
-wizard.
-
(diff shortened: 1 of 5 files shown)


docs/setup_guides/azure.rst
```diff
@@ -1,267 +1,416 @@
...
 
-Using the web interface
-~~~~~~~~~~~~~~~~~~~~~~~
+.. note::
 
-To use the Azure web interface you must first have a Microsoft account that you
-can use to log into the Azure web interface. If you have an existing Microsoft
-account (for example, a hotmail.com or outlook.com email address) then you can use
-that, or you can create a new account.
+   These examples create the Azure Resources in the West Europe region. This is a good
```
TomAugspurger, line 19: I'd feel better with an explicit +1 that it's Ok to include this bit of self-promotion :) I think most pangeo users will be interested in the public data we're hosting there, and so will benefit from knowing about it and putting their compute in the same region as that data.

docs/setup_guides/azure.rst
```diff
@@ -1,267 +1,416 @@
...
-==================================
+We describe two deployments, a :ref:`simple` and a :ref:`recommended`. If you're new to Azure, Kubernetes, or JupyterHub, then you should try the simple deployment to verify that the basics work, before moving on to the more advanced recommend deployment.
 
-The first step to installing Pangeo on Azure is to set up a Kubernetes service
-that can be used to run Pangeo. This can be done either by using the web interface
-or by using the Azure commandline interface (CLI). These are both practical options,
-so we'll cover each one in turn.
+This guide uses the Azure CLI to create the Azure Resources.
+At the end, we provide a ``Makefile`` with targets for the AKS cluster and an updated deployment. If you're just looking to deploy a Hub, feel free to use and adapt the Makefiles. If you're looking to build understanding, read through the guide.
+As an alternative to this guide, you might use `Qhub <https://docs.qhub.dev/en/latest/>`_, which provides a higher-level tool to obtain a JupyterHub and Dask deployment on Kubernetes (or HPC).
```
TomAugspurger, line 15: I'm still exploring QHub. I ran into a couple issues with my initial attempt on Azure, so I've reverted to just using the Azure CLI and Helm in this guide.

choldgraf:
I haven't had a chance to look at the content (and TBH am probably not super qualified to decide haha). But here's my take on a few of your more "meta" points:

- Should Pangeo maintain a guide like this?
  - Ultimately it's the Pangeo community's decision, but my 2 cents is that there is definitely value in having Pangeo-branded guides like this. (with my 2i2c hat on now) While 2i2c will operate the infrastructure for Pangeo, we consider to be operating _Pangeo's stack_ on behalf of Pangeo, not a 2i2c-specific stack. So I'd rather have a community supported "best practices" guide out there that 2i2c can point to and say "2i2c is basically just following these steps".
- Makefiles etc vs step-by-step
  - I think this depends on the kind of value you want these guides to provide to people. With Z2JH we intentionally didn't provide a bunch of "auto deploy" scripts in part because we felt that'd encourage people to just run the script without really learning about the infrastructure. I think that's the danger in the "just copypaste this script or use this abstraction" approach.
- Could this be in Z2JH?
  - This seems super relevant to https://github.com/jupyterhub/team-compass/issues/382. My feeling is that right now, this content belongs outside of Z2JH because there is not yet a cleanly-defined "extension point" for how people could build on the Z2JH docs. But I think it will be a cleaner and better user experience overall if we could move in that direction (not in this PR, but in future ones). 

I also encourage you to update the Z2JH Azure guide if there are now new best practices :-)

sgibson91:
Hi! @TomAugspurger I also probably wrote the networking stuff in the z2jh Azure guide, so very happy to investigate further with you and update as needed (it was a while ago that I wrote it!)

PR #847, opened 2021-07-05
Update ML working group video conference link

djgagne:
Replaced google meet link with zoom to address issues with NCAR host needing to be on the call to let people in.


jhamman:
Thanks @djgagne!

carlosenciso:
Thanks @djgagne

PR #865, opened 2022-02-15
Add support for markdown documents

TomAugspurger:
I haven't translated any, but this adds myst-parser as a dependency to let us use markdown. We can have a mixture of rST and markdown documents.

I think that manual conversion to markdown is best rather than trying to automate it with something like pandoc.

Also fixed all the warnings in the sphinx build.

cc @rsignell-usgs :)

### DIFF
--- a/docs/collaborators.rst
+++ b/docs/collaborators.rst
@@ -32,7 +32,7 @@ Anaconda is funded via a subaward through Columbia.
 We were awarded $1.2M over a three year period (Oct. 2017 - Sept. 2020).
 
 The specific solicitation we responded to is
-`here <https://www.nsf.gov/pubs/2016/nsf16514/nsf16514.htm>`_.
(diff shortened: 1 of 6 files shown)


rsignell-usgs:
Thanks Tom!

rsignell-usgs:
We use "notes" quite a bit in the sphinx-generated documentation.  It's clear how do [notes and warning with restructured text](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html) but can we also do this somehow with Markdown?  (I couldn't find anything googling)

TomAugspurger:
https://myst-parser.readthedocs.io/en/latest/ is the markdown parser. https://myst-parser.readthedocs.io/en/latest/syntax/syntax.html#directives-a-block-level-extension-point documents directives, so this would be the syntax

    ```{note}
    Text of the note
    ```

PR #885, opened 2022-09-27
Minor updates to the speaker links

paigem:
Currently, the links for each speaker's content all point to the main Discourse site, which seems like a broken link. I have made the following changes in this PR:

- Added a link directly to Peter Marsh's talk description on Discourse
- Changed the default link to point to the Pangeo Showcase category on Discourse instead of the main Discourse site (an alternative would be removing the links entirely until there is information posted about that specific talk)
- Fixed a very minor typo

### DIFF
--- a/docs/pangeo-showcase.rst
+++ b/docs/pangeo-showcase.rst
@@ -28,33 +28,33 @@ Fall 2022 Showcase
      - Title
    * - 2022-09-21 12PM EDT
      - Peter Marsh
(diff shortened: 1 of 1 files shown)


JimColl (approved):
Thanks for cleaning those up!  Once this is merged in I'll update with Peter's links.

paigem:
Thanks @JimColl! I'll merge now.

PR #893, opened 2023-02-09
Update showcase schedule

paigem:
Add Campbell Watson to the showcase schedule for May 17th.


JimColl (approved):
Thanks for the updates @paigem

PR #899, opened 2023-03-25
Add Navid to showcase schedule

paigem:
Add Navid to May 10th showcase slot


JimColl (approved):
Thanks for scheduling this!

PR #900, opened 2023-03-29
Updates with new speakers and links

JimColl:
I've added links to all confirmed speakers.


paigem (approved):
Looks great, thanks for this effort @JimColl!

PR #903, opened 2023-04-25
Update COSIMA presentation title

navidcy:


JimColl:
I believe I caught most of this with the recent update, I don't know that the site renders italics so hopefully dropping that extended title is OK with you here?

PR #904, opened 2023-05-09
Add link for COSIMA showcase and removed link for GDN

abarciauskas-bgse:
Removed link for the geospatial discovery network, since I don't see a discourse page for that and I find linking to the showcase page confusing (feels like a broken link)


JimColl (approved):
Thanks @abarciauskas-bgse!

PR #913, opened 2023-08-29
Remove old and unsupported sphinx GHA

ocefpaf:
~~Nothing to see here (yet). Just testing if we can fix the docs build and/or update the sphinx build here.~~

This one is ready for review.


ocefpaf:
@paigem this improves a bit the doc situation here by:

- removing the stale `ammaraskar/sphinx-action` GHA that relies on an [outdated Dockefile](https://github.com/ammaraskar/sphinx-action/pull/21) with old sphinx
- add dependabot for the GHA only, to keep them updated
- install and builds sphinx via the Python GHA + requirements.txt file (easier to maintain, edit, and debug)
- adds taminomara/sphinx-problem-matcher GHA to keep the sphinx inline comments (untested)
- removes [`app.connect("source-read", rstjinja)`](https://github.com/pangeo-data/pangeo/pull/913/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R227) that prevent sphinx from running.

That last step is odd b/c the current approach doesn't error out on it but I could not get an error free build with `rstjinja`. I'm not sure that is actually used so we need input from the folks that implemented that to be sure we can remove it. It was added in https://github.com/pangeo-data/pangeo/pull/293

Or, maybe there is just a malformed jinja somewhere and `rstjinja` is working as expected, not sure.

TL;DR this is not ready for merging :grimacing:

ocefpaf:
OK, the problem was not `rstjinja` but a double documented file, the `hpc.md` and `hpc.rst`. I removed the md version and, if that is OK, this PR is ready for review.

PR #924, opened 2023-12-19
Update Google Analytics Property

### LINKED ISSUE #923: Google Analytics
I'd like to understand traffic to this website.  I asked @rabernat about access to this in checking on that he found that apparently GA is currently broken on this site (maybe the GA-4 migration?).  So now we have two questions:

- Can I get access to GA
- We should fix GA

@rabernat is it only you who has access?  If I can get access then I'll probably want to also grant access to @scharlottej13 who has handled GA-4 transitions before and generally does a great job thinking about traffic and messaging in technical/community situations. Alternatively, if the old GA project isn't broadly and sharing access is hard for some reason  we can also have Sarah create a fresh project and then having her share it with others.

jhamman:
closes #923

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -113,13 +113,14 @@
   </div>
 </footer>
 
-<script async src="https://www.googletagmanager.com/gtag/js?id=UA-120544092-1"></script>
+<!-- Google tag (gtag.js) -->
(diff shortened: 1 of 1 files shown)


jhamman:
This is working now so I'm going to go ahead and merge. As @scharlottej13 suggested in #923, we may be able to leverage Sphinx conf to set this but I think that should be done separately.

